PR #1, opened 2021-01-11
add AGPL-3.0-or-later license

rkusa:
This PR adds the AGPL 3.0 license


LICENSE
```diff
@@ -0,0 +1,661 @@
...
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
rurounijones, line 646: This is the only bit you *really* need in here. You choice if you want to leave the rest in

Update the first line with a description of what this code does and add your copyright date and your name. Take credit 😀

rkusa: Ha! I knew it was a good idea to have a second pair of eyes, I was already wondering why there is no place to put a name and date as in other licenses 😄

rkusa: Interestingly, [most](https://github.com/search?l=&o=desc&q=license%3Aagpl-3.0&s=stars&type=Repositories) seem to copy the whole license. Maybe because creating a PR for adding a license via https://choosealicense.com/licenses/agpl-3.0/ also adds the whole one, but I am in favour of having the excerpt and a link to the whole license, as suggested by "How to Apply These Terms to Your New Programs". Are you fine with he updated authors line?

rurounijones: You have done the vast majority of the work on this so I think sole credit to you as the author is the more accurate choice. However it is up to you. I would squash the commits but again optional.

rkusa: Pity, Github's license detection doesn't work for just having the AGPL excerpt, it needs the whole license. Since I consider his feature useful, because it provides a quick overview of the license ([example](https://github.com/tootsuite/mastodon/blob/master/LICENSE)), I'll probably change it back to the whole license

PR #2, opened 2021-01-15
implement mission event stream

rkusa:
Implements the `StreamEvents` method which returns a stream of `Event`s.

Example:

```bash
$ grpcurl -plaintext -proto ./proto/dcs.proto -d "{}" [::1]:50051 dcs.Mission/StreamEvents
{
  "time": 28804.235,
  "playerEnterUnit": {
    "initiator": "Aerial-1-1"
  }
}
{
  "time": 28804.235,
  "birth": {
    "initiator": "Aerial-1-1"
  }
}
```

I've migrated the event handling from code I already had, so most of the events should work fine. Since there were still minor changes, I am planning to re-test most events to make sure (though there are always edge cases that we'll only catch by actually using the server eventually).

However, the main goal of this PR is to get feedback on the Lua and protobuf code (since I don't have a lot of previous experience with protobuf, so I might for example be doing the whole field numbering wrong) and agree on the RPC interface (naming, data structures, ...) - so feedback (including nitpicking) is welcome!

### DIFF
--- a/proto/dcs.proto
+++ b/proto/dcs.proto
@@ -52,9 +52,9 @@ message Event {
   }
 
   message Position {
-    double x = 1;
-    double y = 2; // alt
-    double z = 3; // y
+    double lat = 1;
+    double lon = 2;
+    double alt = 3; // in meters
   }
 
   // Occurs when a unit fires a weapon (but no machine gun- or autocannon-based
--- a/proto/dcs.proto
+++ b/proto/dcs.proto
@@ -8,9 +8,7 @@ message OutTextRequest {
   bool clear_view = 3;
 }
 
-message OutTextResponse {
-  bool success = 1;
-}
+message OutTextResponse {}
 
 message GetUserFlagRequest {
   string flag = 1;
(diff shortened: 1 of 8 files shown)


proto/dcs.proto
```diff
@@ -8,9 +8,7 @@ message OutTextRequest {
   bool clear_view = 3;
 }
 
-message OutTextResponse {
-  bool success = 1;
-}
+message OutTextResponse {}
```
rkusa, line 11: Is returning at least the `success = true` a best practice, or is just returning nothing also fine?

proto/dcs.proto
```diff
@@ -25,17 +23,270 @@ message SetUserFlagRequest {
...
+  }
+
+  // Occurs when marks get removed.
+  message MarkRemoveEvent {
+    // The unit that removed the mark.
+    string initiator = 1;
+    // The group the mark's visibility is restricted for.
+    uint64 group_id = 2;
+    // The coalition the mark's visibility is restricted for.
+    Coalition coalition = 3;
```
rkusa, line 242: `group_id` and `coalition` aren't always defined, how do we deal with that in protobuf / the RPC. Any ideas?

rkusa: I solved it by using `oneof`, feel free to let me know if you have a better idea.

proto/dcs.proto
rurounijones: Honestly I am not sure. I think nothing is fine since if there was a problem we would be sending an actual error. Based on some reading around of other sources this seems to be fine.

proto/dcs.proto
```diff
@@ -25,17 +23,276 @@ message SetUserFlagRequest {
...
+    oneof target {
+      uint64 id = 2;
+      string name = 3;
+    }
+  }
+
+  message Position {
+    double x = 1;
+    double y = 2; // alt
+    double z = 3; // y
```
rurounijones, line 57: I think it would be better for clients if they only had to deal with decimal Lat/Lon/Alt in a specific format consistently and we handle the conversion internally in the lua using https://wiki.hoggitworld.com/view/DCS_func_LLtoLO and https://wiki.hoggitworld.com/view/DCS_func_LOtoLL

I also don't think it is possible to convert the internal x, y, z information to lat/lon from outside the DCS environment anyway so this becomes basically mandatory.

My suggestions for `Position` would be

```
message Position {
  double lat = 1; // e.g. 41.0000
  double lon = 2; // e.g 37.0000
  double alt = 3;  / e.g. 1000.00 (meters)
}
```

rkusa: Good idea, updated the PR accordingly 👍

PR #3, opened 2021-03-13
Add Unit.getRadar skeleton

rurounijones:
I haven't actually tested that the `getRadar` works but the server
compiles so I wanted to run these changes by you first to see what you
think.

I switched from `proto` directory to `protos` directory since that appeared
to be the standard naming + it is multiple files.

Tested by calling with the ruby client I am working on at the same time.

----------

Add the skeleton code for the dcs `Unit.getRadar` method.

In doing so it became quickly apparent that one .proto file would get
unwieldy quickly so split it out into smaller files. I have created a
common file to hold the enums and objects I expect to be returned from
all manner of methods.

Although I split the .proto file I kept all the methods in the base
`dcs` protobuf package. Although there was a temptation to create
sub-packages I do not believe it would work well when dealing with
methods on objects that inherit since I do not believe the client cares
about which object the method is actually being called on.

This might change as I try to implement some methods from the object
hierarchy so it is best to experiment with this now since this changes
the generated code in both the server and the clients so we want to do
this before things get too painful to change.

### DIFF
--- a/lua/methods/unit.lua
+++ b/lua/methods/unit.lua
@@ -1,0 +1,58 @@
+--
+-- RPC unit actions
+-- https://wiki.hoggitworld.com/view/DCS_Class_Unit
+--
+
+local Unit = Unit
+local Object = Object
+local next = next 
+
+GRPC.methods.getRadar = function(params)
+  local unit = Unit.getByName(params.name)
+  if unit == nil then
+    return GRPC.error("Could not find unit with name '" .. params.name .. "'")
+  end
+    
+  local active, object = unit:getRadar() 
+  
+  if object == nil then
+    env.info("[GRPC]" .. params.name .. " has no radar target")
+    return GRPC.success({   
+      active = active
+    })
+  end
+  
+  env.info("[GRPC]" .. params.name .. "'s Target is " .. object:getCallsign() .. ", ID: " .. object:getID() .. ", Category: " .. object:getCategory() )
+   
+  local category = object:getCategory()
+  local grpc_hash = {}
+  local object_hash = {}
+  
+  if(category == Object.Category.UNIT) then
+    -- TODO, fill out the hash for this object
+    grpc_hash["unit"] = object_hash 
+  elseif(category == Object.Category.WEAPON) then
+    -- TODO, fill out the hash for this object
+    grpc_hash["weapon"] = object_hash 
+  elseif(category == Object.Category.STATIC) then
+    -- TODO, fill out the hash for this object
+    grpc_hash["static"] = object_hash 
+  elseif(category == Object.Category.BASE) then
+    -- TODO, fill out the hash for this object
+    grpc_hash["airbase"] = object_hash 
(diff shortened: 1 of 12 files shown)


rurounijones:
Didn't even see the test section in the code. Hurray for automation!

The problem is obviously that I moved `Position` and `Coalition` messages out of the event scope in the .proto and into the common .proto file and the `dcs` package.

Not sure how to fix it though since my my rust knowledge is 0 and my naive fumbling (removing `event::` or trying `dcs::Position` didn't work).

Any pointers?

rurounijones:
And I have now learned about the `cargo fmt` command 😁

lua/methods/unit.lua
```diff
@@ -0,0 +1,14 @@
+--
+-- RPC unit actions
+-- https://wiki.hoggitworld.com/view/DCS_Class_Unit
+--
+
+GRPC.methods.getRadar = function(params)
+  local unit = Unit:getByName(params.name)
```
rkusa, line 7: Should we return a `GRPC.error("Unit not found")` if `unit` is `nil`?

proto/unit.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message GetRadarResponse {
+  bool active = 1;
+  oneof target {
+    Unit unit = 2;
+    Weapon weapon = 3;
+    StaticObject static_object = 4;
+    SceneryObject scenery_object = 5;
+    Airbase airbase = 6;
+  }
```
rkusa, line 19: To we need a `null` counter-part? I guess that the target is `nil` at least in the case of `active` being `false`?

rkusa (commented):
> I haven't actually tested that the `getRadar` works but the server
> compiles so I wanted to run these changes by you first to see what you
> think.

Looks good 👍, just some small comments

> In doing so it became quickly apparent that one .proto file would get
> unwieldy quickly so split it out into smaller files. I have created a
> common file to hold the enums and objects I expect to be returned from
> all manner of methods.

Totally agree, 👍 for the split

proto/unit.proto
rurounijones: It seems the recommendation and easiest approach is to just not include the field when the value is null so I am doing that.

lua/methods/unit.lua
rurounijones: Done. While looking into this I saw that gRPC has built-in error codes for things like this ( https://grpc.github.io/grpc/cpp/namespacegrpc.html#aff1730578c90160528f6a8d67ef5c43b )

In this case we would want a `NOT_FOUND` code. Would it be possible to modify the `GRPC.Error` command to allow an optional way of setting the error code?

rurounijones:
This latest version is a compiling and working implementation that proves that the serialization of the object types works.

Given the big changes with the proto files I want to merge this asap rather than wait on implementing all the fields in the various objects. I can do that in a later PR.

rkusa (approved):
👍

PR #4, opened 2021-03-22
Add fields for "Unit" object type

rurounijones:
Add fields for the "Unit" object common type and return it as part of
the GetRadar call. This is a test to show that it works in a simple
request / response situation.

Adding the "Unit" object to the EventStream is out of scope for this
commit

---

Tested using the Ruby client

```ruby
client = Dcs::Grpc::Client.new(host: "192.168.1.25")
client.get_unit_radar("Blue Flight-2")
<Dcs::GetRadarRequest: name: "Blue Flight-2">
<Dcs::GetRadarResponse: active: true, unit: <Dcs::Unit: id: "22", name: "Red Flight-2", callsign: "103", coalition: :RED, type: "MiG-29S", position: <Dcs::Position: lat: 42.36043802598069, lon: 42.61754373339121, alt: 1996.1187744140625>>>
```

### DIFF
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -1,0 +1,53 @@
+--
+-- Converts DCS tables in the Object hierarchy into tables suitable for
+-- serialization into GRPC responses
+-- Each exporter has an equivalent .proto Message defined and they must
+-- be kept in sync
+--
+
+local GRPC = GRPC
+local coord = coord
+
+local function toLatLonPosition(pos)
+  local lat, lon, alt = coord.LOtoLL(pos)
+  return {
+    lat = lat,
+    lon = lon,
+    alt = alt,
+  }
+end
+
+GRPC.exporters.unit = function(unit)
+  return {
+    id = unit:getID()
+    name = unit:getName()
+    callsign = unit:getCallsign()
+    coalition = unit:getCoalition();
+    type = unit:getTypeName()
+    position = toLatLonPosition(unit:getPoint())
+  }
+end
+
+GRPC.exporters.weapon = function(weapon)
+  return {}
+end
+
+GRPC.exporters.static = function(static)
+  return {}
+end
+
+GRPC.exporters.airbase = function(airbase)
+  return {}
+end
+
+GRPC.exporters.scenery = function(scenery)
+  return {}
+end
+
+GRPC.exporters.cargo = function(cargo)
+  return {}
+end
+
+GRPC.exporters.object = function(object)
+  return {}
+end
--- a/lua/methods/unit.lua
+++ b/lua/methods/unit.lua
@@ -5,7 +5,7 @@
 
 local Unit = Unit
 local Object = Object
-local next = next 
+local GRPC = GRPC
 
 GRPC.methods.getRadar = function(params)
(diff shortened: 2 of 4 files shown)


lua/methods/unit.lua
```diff
@@ -5,7 +5,7 @@
 
 local Unit = Unit
 local Object = Object
-local next = next 
+local GRPC = GRPC
```
rkusa, line 8: I think I didn't ask in the previous PR, why are the local aliases necessary?

lua/methods/unit.lua
```diff
@@ -16,43 +16,34 @@ GRPC.methods.getRadar = function(params)
...
     return GRPC.success({   
       active = active
     })
   end
-  
-  env.info("[GRPC]" .. params.name .. "'s Target is " .. object:getCallsign() .. ", ID: " .. object:getID() .. ", Category: " .. object:getCategory() )
    
   local category = object:getCategory()
-  local grpc_hash = {}
+  local grpc_table = {}
```
rkusa, line 25: I am wondering whether we should keep it to camelCase in the Lua file? No strong opinion though

protos/common.proto
```diff
@@ -1,46 +1,52 @@
-syntax = "proto3";
```
rkusa, line 1: Looks like you've added a CR to all line endings. Was this on purpose or accidentally (which is probably easy on Windows)? Might be worth adding `core.autocrlf true` a local Git config, wdyt?

lua/exporters/object.lua
```diff
@@ -0,0 +1,54 @@
...
+end
+
+GRPC.exporters.unit = function(unit)
+  local u = {}
+  u["id"] = unit:getID()
+  u["name"] = unit:getName()
+  u["callsign"] = unit:getCallsign()
+  u["coalition"] = unit:getCoalition();
+  u["type"] = unit:getTypeName()
+  u["position"] = toLatLonPosition(unit:getPoint())
```
rkusa, line 27: Any reason to not assign anything at once, ala:

```lua
local u = {
  id = unit:getID(),
  ...
}
```

Or mainly personal preference? I don't have a strong opinion here either, so the question is mostly about being curious.

rkusa (commented):
Good stuff, just some minor things 🙂

lua/methods/unit.lua
rurounijones: From what I have read snake_case and camelCase is personal preference. I used snake_case because ruby influence but will switch to camelCase because that is what the ED APIs use.

lua/exporters/object.lua
rurounijones: No proper reason. I am still fumbling around in lua (god I do not enjoy writing lua...)

Switched to the suggested layout.

lua/methods/unit.lua
rurounijones: I am completely cargo-culting this but from what I have read it improves performance.

protos/common.proto
rurounijones: Completely accidental. Added the config

rkusa (approved):
👍

PR #6, opened 2021-04-03
Add GetAirbases API

rurounijones:
Add API to retrieve airbase information. This includes things like
carriers and FARPS.

### DIFF
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -19,11 +19,11 @@ end
 
 GRPC.exporters.unit = function(unit)
   return {
-    id = unit:getID()
-    name = unit:getName()
-    callsign = unit:getCallsign()
-    coalition = unit:getCoalition();
-    type = unit:getTypeName()
+    id = unit:getID(),
+    name = unit:getName(),
+    callsign = unit:getCallsign(),
+    coalition = unit:getCoalition(),
+    type = unit:getTypeName(),
     position = toLatLonPosition(unit:getPoint())
   }
 end
@@ -37,7 +37,20 @@ GRPC.exporters.static = function(static)
 end
 
 GRPC.exporters.airbase = function(airbase)
-  return {}
+  local a = {
+    name = airbase:getName(),
+    callsign = airbase:getCallsign(),
+    coalition = airbase:getCoalition(),
+    category = airbase:getDesc()['category'],
+    displayName = airbase:getDesc()['displayName'],
+    position = toLatLonPosition(airbase:getPoint()),
+  }
+
+  if airbase:getUnit() then
+    a.id = airbase:getUnit():getID()
+  end
+
+  return a
 end
 
 GRPC.exporters.scenery = function(scenery)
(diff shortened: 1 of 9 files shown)


lua/exporters/object.lua
```diff
@@ -19,11 +19,11 @@ end
...
-    id = unit:getID()
-    name = unit:getName()
-    callsign = unit:getCallsign()
-    coalition = unit:getCoalition();
-    type = unit:getTypeName()
+    id = unit:getID(),
+    name = unit:getName(),
+    callsign = unit:getCallsign(),
+    coalition = unit:getCoalition(),
+    type = unit:getTypeName(),
```
rurounijones, line 26: Added the commas for consistency with usage elsewhere

protos/common.proto
```diff
@@ -47,6 +53,13 @@ message Scenery {
 }
 
 message Airbase {
+	string id = 1;
```
rurounijones, line 56: This doesn't actually work yet. The rust serializer errors out expecting that Id *must* be present

`021-04-02 16:09:24.797 ERROR   SCRIPTING: [GRPC] Error retrieving next command: Failed to deserialize result for method getAirbases: deserialize error: missing field id`

 which seems to be a bug based on https://github.com/danburkert/prost/issues/430 and the comment

> As the README notes, proto3 message fields without a modifier are be wrapped in an Option. The new proto3 optional modifier is for scalar fields, I'm kind of surprised protoc even lets you apply it to a message field.

As a test I did try to apply `optional` to this field but, as expected, this fails.

`protoc failed: common.proto:56:18: Explicit \'optional\' labels are disallowed in the Proto3 syntax. To define \'optional\' fields in Proto3, simply remove the \'optional\' label, as fields are \'optional\' by default`

rurounijones (commented):
This is not ready to merge yet until the `optional` / not `optional` thing is figured out. See review comments for details

src/rpc.rs
```diff
@@ -82,6 +82,18 @@ impl Mission for RPC {
...
 
+    async fn get_airbases(
+        &self,
+        request: Request<GetAirbasesRequest>,
+    ) -> Result<Response<GetAirbasesResponse>, Status> {
+        let res: GetAirbasesResponse = self
+            .ipc
+            .request("getAirbases", Some(request.into_inner()))
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
```
rkusa, line 93: I've simplified the API here part a bit, so we can shorten it to the following:

```suggestion
        let res: GetAirbasesResponse = self.request("getAirbases", request).await?;
```

rkusa: I made the change, I hope you don't mind

rkusa (approved):

src/rpc.rs
rurounijones: Awesome, thanks!

rurounijones:
Thanks for investigating and fixing.

I need to do some more testing with the various things that qualify as `airfields` (Aircraft carriers, helo carriers, FARPS, oilrigs etc) and will merge if that is successful.

rurounijones:
Fully implemented the API allowing for optional specifying of the coalition Id. I am going for optional here since it makes more sense than adding another enum value for `all` which could bite us later.

PR #7, opened 2021-06-12
Add DCT functions

rurounijones:
Add some DCT functions to allow for use on DCT framework using servers.

See commit message for implementation details. I am not sure what the best way to organise the .protos is yet so going for the simplest option (just adding a "custom" category) to make it easier to reorganise later when we have more APIs implemented which might suggest a better organisation scheme.

The general idea of these is that we can use DCS-gRPC to provide the generic'ish RPC methods but in the cases where we are not directly implementing a built-in ED DCS method we leave them unimplemented so that the servers can implement the methods how they like.

In this case the methods are implemented by the DCT framework in https://github.com/ricmzn/dct/blob/syria-at-war/src/dct/systems/overlordBotRPC.lua

I have updated the C# client with these definitions: https://github.com/DCS-gRPC/csharp-client

And an implementation in OverlordBot: https://gitlab.com/overlord-bot/srs-bot/-/commit/a6bfae4af0c878c4317b49849f85feb5f57244cf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 !.github
 /target
 Cargo.lock
+scratchpad.txt
+grpcurl.exe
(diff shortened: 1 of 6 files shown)


rkusa (approved):
LGTM 👍

Issue #8, opened 2021-06-13
Stream basic unit information (name, type, position)


rurounijones:
EDIT

I think the easiest thing to do so far is to just use the MIST code where the lua will literally just be reading the `mist.dbs.aliveUnits` table, maybe converting them using the existing exporters and returning the table with the exported versions.

The question is how to have the rust code call this method on a timer and stream the objects to all clients. Since this is rust area I have very little idea of what needs to be done

---

Putting this in a Github Issue for discussion and brain storming.

We would like to export unit position information as a gRPC stream to clients. We want to do this in a way that does not block the lua thread in large missions. We want to do it at a use specified refresh rate which would be coarse to the level of something like every 1 second at highest refresh rate but could be as slow as every 10 seconds.

Looking for ideas since some of this will require more rust coding than just acting as an RPC interface I think (The perform in chunks option especially).

The basic process is: 

1. Client calls the RPC with desired refresh rate
2. Call [coalition.getGroups](https://wiki.hoggitworld.com/view/DCS_func_getGroups) for coalition 1.
3. For each group, call [getUnits](https://wiki.hoggitworld.com/view/DCS_func_getUnits) and add to a lua table
4. Return the lua table to the rust code that streams each entry after [serializing](https://github.com/DCS-gRPC/rust-server/blob/main/lua/exporters/object.lua) it
5. IF coalition == 1 then GOTO 2 for coalition 2
5. Sleep until refresh time and GOTO 2

This will work but it will get *all* units for each coalition in one go which may end up being many units and that could block the lua thread.

Possible ways around this....

# Perform in chunks

1. Client calls the RPC with desired refresh rate
2. Call `coalition.getGroups` for coalitions 0, 1, 2 and add to a united `groups` table which is returned to rust
4. For each group call `Group.getUnits` and return a lua table that contains the units to the rust code that iterates and serializes and streams each unit.
5. Sleep TIME_REMAINING_UNTIL_REFRESH / GROUPS_REMAINING in rust.
6. IF GROUPS_REMAINING > 1 GOTO 2 ELSE SLEEP  TIME_REMAINING_UNTIL_REFRESH THEN GOTO 2

# Use lua coroutines

This is the same as the basic process but we use lua coroutines and add some `yields` in the code when it iterates groups (or even units). I have no idea if coroutines will work as expected when the method is being called from rust or even if they will work properly.

[...]

PR #11, opened 2021-07-09
Add `GetGroups`, `GetUnits` and `GetUnitPosition` methods

rkusa:
First set of methods to support #8 eventually.

This PR adds the following methods:

- `GetGroups`

    ```
    grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"coalition\": 2}' 127.0.0.1:50051 dcs.Mission/GetGroups
    ```

- `GetUnits`

    ```
    grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"groupName\": \"Aerial-1\"}' 127.0.0.1:50051 dcs.Mission/GetUnits
    ```

- `GetUnitPosition`

    ```
    grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"name\": \"Aerial-1-1\"}' 127.0.0.1:50051 dcs.Mission/GetUnitPosition
    ```

- `StreamUnits`

    ```
    .\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{}' 127.0.0.1:50051 dcs.Mission/StreamUnits
    ```


protos/common.proto
```diff
@@ -18,6 +18,14 @@ enum AirbaseCategory {
   SHIP = 2;
 }
 
+enum GroupCategory {
+  GROUP_AIRPLANE = 0;
+  GROUP_HELICOPTER = 1;
+  GROUP_GROUND = 2;
+  GROUP_SHIP = 3;
+  GROUP_TRAIN = 4;
```
rkusa, line 26: Prefix is apparently necessary if multiple enums have the same variant, see the error I am getting otherwise:

```
\"SHIP\" is already defined in \"dcs\".\r\ncommon.proto:25:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  
Therefore, \"SHIP\" must be unique within \"dcs\", not just within \"GroupCategory\"
```

protos/common.proto
```diff
@@ -40,8 +48,12 @@ message Unit {
   string callsign = 3;
   Coalition coalition = 4;
   string type = 5;
-  Position position = 6;
```
rkusa, line 43: I was wondering whether we should only provide non-static values via dedicated RPC calls. Which would be `GetUnitPosition` in this case.

protos/group.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+import "common.proto";
+
+package dcs;
+
+message GetUnitsRequest {
+  string group_name = 2;
```
rkusa, line 8: I am not sure about the inconsistency between `GetUnitsRequest` and `GetUnitPositionRequest `, where this one here uses `group_name` while the latter uses `name`. I wen with this for now, because the latter has `Unit` already in its method name.

protos/common.proto
rurounijones: I think that is fine, we may want to break down the gRPC namespace later but that is a problem for future, if it even becomes needed.

protos/common.proto
rurounijones: Is the thinking for this due to performance reasons? I am honestly not sure how much extra time it takes to return a full unit object vs just the position but I am not sure it would be worth separate calls..

rurounijones: I personally don't really see a need for a separate "position" call but that is based on my thinking that the amount of time it takes to create a position response vs a full Unit response object is negligable.

However I don't mind having a separate request _in addition_ to including the position in a full Unit response object. But I am not in favour of _requiring_ a separate call.

protos/group.proto
rurounijones: Would it be clearer to name this `GetUnitsInGroup` which would make it clearer why this is `group_name` but the position would be `name` (since it is more obviously for a specific unit).

lua/methods/group.lua
```diff
@@ -0,0 +1,24 @@
...
+
+local GRPC = GRPC
+
+GRPC.methods.getUnits = function(params)
+  -- https://wiki.hoggitworld.com/view/DCS_func_getByName
+  local group = Group.getByName(params.groupName)
+  if group == nil then
+    return GRPC.errorNotFound("group does not exist")
+  end
+
```
rurounijones, line 14: Just want to confirm. There is a race condition here if the group gets removed before we call `getUnits`.

I am personally fine with it since I think the changes of that actually happening are infintessimal. Just want to make sure you are also happy with it.

protos/common.proto
rurounijones: I think this change is fine, we may want to break down the gRPC namespace later to avoid this but that is a problem for future (If it is even a real problem).

I am not familiar enough with gRPC in rust, C# and ruby (the languages currently in use) to evaluate the pros and cons so lets go with the above since it is the simplest approach.

protos/common.proto
rurounijones: I personally don't really see a need for a separate "position" call but that is based on my thinking that the amount of time it takes to create a position response vs a full Unit response object is negligable.

However I don't mind having a separate request in _addition_ to including the position in a full Unit response object since it lets the client decide which they needed. But I am not in favour of _requiring_ a separate call.

Do we need a generic `GetUnit` call as well? Right now we can get the units in a group and get the units position but cannot get the full info on an individual unit.

protos/group.proto
rurounijones: Would it be clearer to name this GetUnitsInGroup which would make it clearer why this is group_name but the position would be name (since it is more obviously for a specific unit).

rurounijones (commented):
Yaaay! Thanks for the changes 😁 

You say in your comments that this is a pre-requisite for #8 but could you explain how? I think there is some rustiness nuance bit here that I am missing.

My current understanding is that gRPC methods are purely called by network clients but are you planning on calling these from inside the Rust server itself to implement the streaming or something along those lines?

rkusa:
Thanks for the feedback!

> You say in your comments that this is a pre-requisite for #8 but could you explain how? I think there is some rustiness nuance bit here that I am missing.
> 
> My current understanding is that gRPC methods are purely called by network clients but are you planning on calling these from inside the Rust server itself to implement the streaming or something along those lines?

I am not sure if I am going to re-use the gRPC methods directly, but I am definitely planning on re-using the Lua part and maybe the gRPC request and response structs. The rough plan I have in the back of my head (not thought out completely yet, and might still change):

- Create a new gRPC service with a method to get all alive units. I am not sure about the name yet, but I am thinking about naming the service in a way that makes it clear that this is not simply forwarding calls to the mission env and instead does some extra internal stuff. Something in the direction of `dcs.Extra/GetUnits` (I don't like this name, but should show the direction I am thinking about).
- Once this method is first called, it does an initial sync using `self.request("getGroups", request).await` and `self.request("getUnits", request).await`
- After the initial sync:
  - it always responds with its internal _copy_ of alive units when `dcs.Extra/GetUnits` is called
  - it tries to keep its internal copy up to date by listening to events via `self.ipc.events().await`
  - since the above is not the most reliable method in DCS, it does some extra checks per unit on a regular basis

Does this make sense to you, or do you see any issues or is it maybe even going into the wrong direction for what you were aiming for in #8?

protos/common.proto
rkusa: While I would probably tend to only expose the position via an additional RPC call in a Rust client, having the position in both the full Unit response object and the separate call is actually still fine for me. Could probably be useful in context of an event to know the position at the time of the event.

```rust
struct Unit {
  position: Position,
}

impl Unit {
  fn get_position(&self) -> Position {
    // what to return? The potentially outdated `self.position`, or the result of an RPC call to `GetUnitPosition`?
  }
}
```

The reason I'd prefer the separate call in a client is that I wouldn't know what to return when the client user asks for the position. Should I get the most recent position or return a potentially outdated one?

> Do we need a generic `GetUnit` call as well? Right now we can get the units in a group and get the units position but cannot get the full info on an individual unit. (Although that might not be needed for this so I can add that later)

I'd say no, mainly because I am not aware of a DCS Lua method to get a specific unit by name (we can only get units by going through a group AFAIK).

The question is also, what is the full response? I'd tend towards requiring a separate RPC call for all information that might change like position, health, ammo etc. instead of one that returns all of them.

lua/methods/group.lua
rkusa: Good question, I am somehow in the assumption that the lua blocks the whole mission simulation so that this case would be impossible. But I am just assuming, I don't know if it is really this case

rurounijones:
Ok, so I _think_ I understand what you are going for but I am not sure how useful it would be to a client to have units without the curent position information.

My vision is that the client calls `Mission.StreamUnits` or something like that and receives a constant stream of up-to-date unit objects (That includes their current position).

The stream would either be individual unit objects sent at a fixed rate or a single array of all the units in the server every second (for example) or streaming an array of units in a group at a time .

This way the client just calls one API and gets a constant stream of updates. This removes the need for having to to have a separate "GetUnitPosition" API that the client would then need to call separately. In the use-cases I can think of, such as OverlordBot or a WebGCI page, the client would have to be calling the position update constantly _anyway_

The implications of this are that the Rust server needs to call into the mission environment to keep the unit positions up to date so maybe it is just as fast to call into the mission env and get all the unit information rather than keeping a separate copy around inside rust and keeping that updated with the current position.

In order to not stop-the-world by getting every unit in the game at once we could do something like 1 getGroup call then every $X_TIME we get all the units in a group and stream those units before moving on to the next group.  I believe that is what LotATC does. To use some pseudocode here

```ruby
while shutdown == false
  groups = missionenv.getGroups(coalition)
  groups.each do |group|
    units = missionenv.getUnitsInGroup(group)
    streamUnitsToClient(units)
    sleep(X)
  end
end  
```

The X could be configurable such that we should cycle through every group in a 2 second period so 2 / groupCount or something like that.

rkusa:
Rebased onto `main` (due to merge conflicts). I'll add an initial version of the unit streaming before moving the PR out of its draft state again

lua/exporters/object.lua
```diff
@@ -8,32 +8,32 @@
...
 local coord = coord
 
-local function toLatLonPosition(pos)
-  local lat, lon, alt = coord.LOtoLL(pos)
-  return {
-    lat = lat,
-    lon = lon,
-    alt = alt,
-  }
-end
```
rkusa, line 18: This already existed as `GRPC.toLatLonPosition`

lua/grpc.lua
```diff
@@ -78,13 +78,14 @@ end
...
 dofile(GRPC.basePath .. [[methods\atmosphere.lua]])
+dofile(GRPC.basePath .. [[methods\coalitions.lua]])
+dofile(GRPC.basePath .. [[methods\custom.lua]])
+dofile(GRPC.basePath .. [[methods\group.lua]])
+dofile(GRPC.basePath .. [[methods\mission.lua]])
 dofile(GRPC.basePath .. [[methods\trigger.lua]])
 dofile(GRPC.basePath .. [[methods\unit.lua]])
 dofile(GRPC.basePath .. [[methods\world.lua]])
-dofile(GRPC.basePath .. [[methods\custom.lua]])
-dofile(GRPC.basePath .. [[methods\event_stream.lua]])
```
rkusa, line 87: Just sorted alphabetically, because I couldn't figure out where to put my new file :D

lua/methods/coalitions.lua
```diff
@@ -15,5 +15,20 @@ GRPC.methods.getPlayers = function(params)
...
   end
   return GRPC.success({units = result})
+end
 
-end
\ No newline at end of file
+GRPC.methods.getGroups = function(params)
+  local result = {}
+  for _, c in pairs(coalition.side) do
+    if params.coalition == nil or params.coalition == c then
```
rkusa, line 23: Made `params.coalition` optional

lua/methods/unit.lua
```diff
@@ -45,4 +45,30 @@ GRPC.methods.getRadar = function(params)
...
+
+GRPC.methods.getUnitPlayerName = function(params)
+  -- https://wiki.hoggitworld.com/view/DCS_func_getByName
+  local unit = Unit.getByName(params.name)
+  if unit == nil then
+    return GRPC.errorNotFound("unit does not exist")
+  end
+
+  return GRPC.success({
+    -- https://wiki.hoggitworld.com/view/DCS_func_getPoint
```
rkusa, line 71: - [ ] Need to update the link

protos/dcs.proto
```diff
@@ -1,13 +1,14 @@
...
+import "common.proto";
+import "custom.proto";
+import "group.proto";
+import "mission.proto";
 import "trigger.proto";
 import "unit.proto";
 import "world.proto";
-import "custom.proto";
-import "common.proto";
-import "coalitions.proto";
```
rkusa, line 10: Sorted by name too 🙈

protos/mission.proto
```diff
@@ -232,4 +232,33 @@ message Event {
...
+  // to the DCS mission.
+  // Default: 5
+  optional uint32 poll_rate = 1;
+
+  // The maximum backoff in seconds which the gRPC postpones polling units that haven't moved
+  // recently. This is an optimization to dynamically reduce the poll rate for stationary units.
+  // Set it to the same value as `poll_rate` to disable the backoff.
+  // Default: 30
+  optional uint32 max_backoff = 2;
+}
```
rkusa, line 249: Wdyt, should we expose those options or be opinionated about them instead?

protos/mission.proto
```diff
@@ -232,4 +232,33 @@ message Event {
...
+  }
+
+  oneof update {
+    // The unit is either new or its position changed.
+    Unit unit = 1;
+
+    // The unit does not exist anymore.
+    UnitGone gone = 2;
+  }
+}
```
rkusa, line 264: Let me know if you have a better idea of how to communicate unit updates vs an information that a unit does not exist anymore.

src/stream.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        Ok(changed)
+    }
+}
+
+/// Check whether two positions are equal, taking an epsilon into account.
+fn eq(a: &Position, b: &Position) -> bool {
+    const LL_EPSILON: f64 = 0.000001;
+    const ALT_EPSILON: f64 = 0.001;
```
rkusa, line 280: Wdyt about the `Position` comparison here. Should we myb increase or decrease the precision of the comparison? I think for the altitude an epsilon of `0.01` might be perfectly fine too.

src/stream.rs
```diff
@@ -0,0 +1,293 @@
...
+            // poll units for updates
+            _ = interval.tick() => {
+                update_units(&mut state).await?;
+            }
+        }
+    }
+}
+
+/// The state of an active units stream.
+// TODO: re-use one state for all concurrent unit streams?
```
rkusa, line 101: I tend towards keeping a separate state per unit stream to not throttle another stream because one of the clients is very slow in consuming its data.

rurounijones: > I tend towards keeping a separate state per unit stream to not throttle another stream because one of the clients is very slow in consuming its data.

Are there any advantages to having an mpmc stream? Would it result in less load on the server or something like that? I agree with yout bias of not letting one client screw the others via an accidental "slowloris" attack.

When I wrote tac_scribe, that reads tacview events, I made sure that one thread's job was to receive messages and put them on an internal memory queue for actual processing. This made sure that the receiving thread was fast and would not cause any slowdown.

If there are serious advantages to having mpcp stream then I think we might consider using it and documenting the crap out of the stream system and make sure clients do something similar to tac_scribe but ultimately I am not too fussed either way and defer to you on this one.

protos/mission.proto
rurounijones: What you have chosen above is basically how Tacview does it, I think it works.

protos/mission.proto
rurounijones: I am fine either way. At least initially while we are testing, we may decide to make it opinionated later? (Or at least have minimum values)

src/stream.rs
rurounijones: I will do some testing this weekend and see if there is anything to suggest that we need to tune these values.

lua/exporters/object.lua
rurounijones: Whoops, understood

lua/grpc.lua
rurounijones: Good catch.

lua/methods/coalitions.lua
rurounijones: It seems wierd that we have a `coalition` service method that does not take a coalition parameter. We are going to break from DCS standards whatever we do here so should we stick with this or maybe move this somewhere else?

rurounijones:
Tested over the weekend on a light mission with units getting spawned and destroyed. Seemed to work without any issues.

Is there any way we can measure time spent in lua or something like that to reassure server owners that it is not very impactful? (Either some test or via logging performance metrics?)

Jupiter2

rurounijones:
After doing some more UI work there are some things we need to think about how to do.

### Group ID

There is no way to associate a unit with a group from unit streaming. This is because the Unit does not include the group information or even a group Id.

This makes streaming much more difficult (especially f you are trying to build a treeview for units for example)

#### Include the Group 

We could always include the group object (ID and name only, not the units) (using `getGroup()` on the unit) or, if we have the info available, only include the group the first time a unit is seen.

#### Include the Group ID

Only include the groupId in the Unit object. This will give us enough information to create a group in our UI with a unique ID and the rest of the information for that group can be included in a `GetGroups` call (but in that call we would not want the units, so we would need a "GetGroupsWithoutTheUnits" call or option paramter which doesn't sit well).

#### Include the group ID and the group name

Include the group ID and the group name in the Unit so that we have all the information we need for building a UI up-front.

#### Stream Group objects with name and ID only.

Stream group objects with ID and name only as part of the the streaming data alongside `Unit` and `UnitGone`

#### Stream group objects with *everything* upon first load.

The first time a new group is seen then stream the group object with all its associated units. This would require an internal rust list of "known" groups per client so that it knows what to send

-----------

Not sure what solution is the best here. They all have their pros and cons. What do yout think?

protos/group.proto
rkusa: I like the suggestion of using namespaces here. I've moved it into a namespace in here https://github.com/DCS-gRPC/rust-server/pull/11/commits/81a5bd7ddda73731d17e712d56e6f7efae15782c, to see what it would look like. I like it, wdyt? Though it probably means that we should also add namespaces to all the other types for consistency reasons later on.

lua/methods/coalitions.lua
rkusa: I agree and updated the implementation in here a5f0eebd40faf87c2885600aa21a7016d963798e

lua/methods/unit.lua
rkusa: Done in af9380fee85cda81073b631f6b49432afbbe60f7

rkusa:
> Not sure what solution is the best here. They all have their pros and cons. What do yout think? I think just including the GroupId in the unit object is probably the easiest and most consistent-with-other-things option.
> 
> I did the modifications locally and got the following. The idea is that a "new group" is created each time a unit appears with a new group Id. And then in another thread we will call `GetGroups` on a timer to then populate the name of the group separately.

I'd probably go for including the group name for each unit. I tend towards the name because there is a [`getByName`](https://wiki.hoggitworld.com/view/DCS_func_getByName), but there is no `getByID`. Wdyt about name instead of id?

src/stream.rs
rkusa: I gave it some addition thought and I think the question comes down to:

a) Should we expose unit-stream-instance-specific settings (see https://github.com/DCS-gRPC/rust-server/pull/11#discussion_r679436403) or
b) not (= shared state).

When not (b), we can re-use the same state and reduce the amount of calls made to the mission environment in case of concurrent unit streams. I think I could also make it so that one client cannot negatively affect each other (by handling updates very slow), but I'd probably have to buffer updates for each clients and thus terminate them if their respective buffer gets full.

protos/group.proto
rurounijones: Adding everything into namspaces is part of the refactoring I want to do once this PR is merged.

rurounijones:
Whoops, yes, `GroupName` is the better option.

If we are using `GroupName` instead of the `GroupId` in the unit then the first unit that appears with a so-far-unseen group name can automatically create the group as well.

This means that streaming `Group` objects becomes unecessary for the purposes of the UI I posted above. I am trying to think of any scenario where streaming group objects would be beneficial but cannot think of anything right now. Since GroupNames have to be unique anyway it isn't like we are losing anything by not having an ID.

protos/common.proto
rurounijones: I agree that we don't want a call that returns *everything* for a unit but position is the exception. Unit position It is so integral to every use-case I can think of for unit streaming use-cases that would use this that it should be included.

If a client specifically calls the `get_position` function then I do think that it should return the results of a `GetUnitPosition` call and not `self.position`. If a client is specifically asking for the position of a unit then chances are it is not actually using the streaming API at all otherwise it would already have that info.

Although in all honesty, I cannot really think of a situation where manually getting groups, units and unit positions is more efficient than just receiving the data via the streaming API so we might be in a case of discussing something that is YAGNI (You ain't gonna need it) anyway.

src/stream.rs
rurounijones: I am all for the option that reduces the number of calls made into the mission environment (Which I think is priority #1) and I don't see a problem with re-using the state.

I think the idea of having a buffer and auto-termination is a good idea. One potential issue I can see with auto-termination is that the queue could get instantly full on the initial unit dump as the client connects and therefore the client gets instantly disconnected? Not sure.

rkusa:

rkusa:
We've decided to merge the PR as is and create follow-ups for the remaining open points

PR #12, opened 2021-07-09
Multiplex separate gRPC services to one socket

rurounijones:
Multiplex multiple gRPC services into one socket so that we can more
rationally organise the APIs rather than have everything under
`dcs.Mission`.

Eventually we will split out the services further. This commit is the
proof of concept.

Had to move the `add_service` code out of the builder because it
required a private module in the signature. Since I still have 0
knowledge of Rist please let me know what improvements I can
do. I just wanted to get the code working.

Tested on a server and verified that the Streaming and OutText
commands still worked and it _appears_ to but I noticed twice
that the streaming froze until I closed the client connection so
I am suspicious that these changes have introduced some
non 100% replicatable issue.


src/server.rs
```diff
@@ -22,13 +25,19 @@ pub async fn run(ipc: IPC<Event>, mut shutdown_signal: Receiver<()>) {
...
 
     let addr = "0.0.0.0:50051".parse().unwrap();
-    RPC::builder(ipc)
+    Server::builder()
+        .add_service(MissionServer::new(RPC {
+            ipc: ipc_event.clone(),
+        }))
+        .add_service(StreamerServer::new(RPC {
+            ipc: ipc_event.clone(),
+        }))
```
rkusa, line 40: We can re-use the same RPC here

```suggestion
    let rpc = RPC::new(ipc);
    Server::builder()
        .add_service(MissionServer::new(rpc.clone()))
        .add_service(StreamerServer::new(rpc))
```

src/server.rs
```diff
@@ -22,13 +25,19 @@ pub async fn run(ipc: IPC<Event>, mut shutdown_signal: Receiver<()>) {
 }
 
 async fn try_run(
-    ipc: IPC<Event>,
+    ipc_event: IPC<Event>,
```
rkusa, line 28: Nit: I personally don't see a benefit of the added `_event` suffix (but am also not opposed to it)

src/rpc.rs
```diff
@@ -1,28 +1,21 @@
...
-    ipc: IPC<Event>,
+    pub ipc: IPC<Event>,
 }
 
 impl RPC {
-    pub fn builder(
-        ipc: IPC<Event>,
-    ) -> Router<MissionServer<RPC>, transport::server::Unimplemented> {
-        Server::builder().add_service(MissionServer::new(RPC { ipc }))
-    }
```
rkusa, line 24: A common approach in Rust here would be to use a `new` _constructor_:

```rust
    pub fn new(ipc: IPC<Event>) -> Self {
        RPC { ipc }
    }
```

protos/dcs_mission.proto
```diff
@@ -29,6 +29,15 @@ service Mission {
 
   // DCT Function
   rpc JoinMission(MissionJoinRequest) returns (MissionJoinResponse) {}
+}
 
-  rpc StreamEvents(StreamEventsRequest) returns (stream Event) {}
+service Streamer {
+	// Streams DCS game generated Events.
+	// See https://wiki.hoggitworld.com/view/Category:Events
+	rpc Events(StreamEventsRequest) returns (stream Event) {}
```
rkusa, line 37: I'd actually tend towards keeping `Events` inside of the `Mission` service, and split the services by e.g. `Mission`, `Unit`, `Group`, ...

src/rpc.rs
```diff
@@ -1,28 +1,21 @@
...
-use tonic::transport::{self, Server};
 use tonic::{Request, Response, Status};
 
 pub mod dcs {
     tonic::include_proto!("dcs");
 }
 
 pub struct RPC {
-    ipc: IPC<Event>,
+    pub ipc: IPC<Event>,
```
rkusa, line 15:
```suggestion
    ipc: IPC<Event>,
```

In case you add the `new` constructor suggested in the next comment

rkusa (commented):
Thanks for starting the split 👍 

> Tested on a server and verified that the Streaming and OutText
> commands still worked and it _appears_ to but I noticed twice
> that the streaming froze until I closed the client connection so
> I am suspicious that these changes have introduced some
> non 100% replicatable issue.

I couldn't see any issue (neither in the code nor while testing), might be something else.

protos/dcs_mission.proto
rurounijones: That is fine, this was just an easy one to test with.

src/server.rs
rurounijones: Done

src/server.rs
rurounijones: I had to do it because otherwise the program was throwing errors. However this is no longer needed due to other suggested changes.

src/rpc.rs
rurounijones: Done

protos/dcs_mission.proto
rurounijones: Split more fully, as discussed in Discord we will make the services plural (Mission will stay singular).

src/rpc.rs
rurounijones: Done

README.md
```diff
@@ -51,7 +51,7 @@ dofile(GRPC.basePath .. [[grpc.lua]])
 - Seach for `[GRPC]` in the DCS logs
 - Consult the gRPC Server logs at `Saved Games\DCS.openbeta\Logs\gRPC.log`
 
-Test the running server via [grpcurl](https://github.com/fullstorydev/grpcurl):
+Test the running server via [grpcurl](https://github.com/fullstorydev/grpcurl): (Remove the `.exe` when running on Linux)
```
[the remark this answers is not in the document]
rurounijones: No, I am using windows and I assume most others will as well.

rurounijones:
I tried rebasing on `main` but was having some issues I could not figure out.

rkusa:
Rebased onto `main` (due to merge conflicts)

PR #13, opened 2021-07-23
Add Atmosphere Query APIs

rurounijones:
Implement the 3 atmospheric APIs provided by the mission environment. I am sticking with:
* The metric units for the wind strength
* Using the western convention of the wind heading being where it is coming from, not where it is heading to.
* The DCS provided values for temperature and pressure (Kelvin and Pascals)

I am also not doing any rounding. Let the clients determine what level of accuracy they want.

Tested using grpcurl.

```plain
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"position\": { \"lat\": 43.116012753760046, \"lon\": 41.96052007622479, \"alt\": 6089.67431640625 }}' 127.0.0.1:50051 dcs.Atmosphere/GetWind
{
  "direction": 240.22713,
  "speed": 9.941755
}                       
```
```plain
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"position\": { \"lat\": 43.116012753760046, \"lon\": 41.96052007622479, \"alt\": 6089.67431640625 }}' 127.0.0.1:50051 dcs.Atmosphere/GetWindWithTurbulence
{
  "direction": 239.76268,
  "speed": 9.890662
}
```
```plain                                                                                                                                                                                                                   
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"position\": { \"lat\": 43.116012753760046, \"lon\": 41.96052007622479, \"alt\": 6089.67431640625 }}' 127.0.0.1:50051 dcs.Atmosphere/GetTemperatureAndPressure
{
  "temperature": 253.56712,
  "pressure": 47322.81
}
```

### DIFF
--- a/lua/methods/atmosphere.lua
+++ b/lua/methods/atmosphere.lua
@@ -1,0 +1,56 @@
+--
+-- RPC atmosphere actions
+-- https://wiki.hoggitworld.com/view/DCS_singleton_atmosphere
+--
+
+local atmosphere = atmosphere
+local coord = coord
+local GRPC = GRPC
+
+local function convertWindResults(windvec3)
+  local direction = math.deg(math.atan2(windvec3.z, windvec3.x))
+  
+  if direction < 0 then
+    direction = direction + 360
+  end
+  
+  -- Convert TO direction to FROM direction. 
+  if direction > 180 then
+    direction = direction-180
+  else
+    direction = direction+180
+  end
+  
+  -- Calc 2D strength.
+  local strength = math.sqrt((windvec3.x)^2+(windvec3.z)^2)
+
+  return  {
+    heading = direction, -- Western style. Heading the wind is coming FROM
+    strength = strength -- m/s
+  }
+end
+
+GRPC.methods.getWind = function(params)
+  local point = coord.LLtoLO(params.position.lat, params.position.lon, params.position.alt)
+     
+  return GRPC.success(convertWindResults(atmosphere.getWind(point)))
+end
+
+GRPC.methods.getWindWithTurbulence = function(params)
+  local point = coord.LLtoLO(params.position.lat, params.position.lon, params.position.alt)
+     
+  return GRPC.success(convertWindResults(atmosphere.getWindWithTurbulence(point)))
+end
+
+GRPC.methods.getTemperatureAndPressure = function(params)
+  local point = coord.LLtoLO(params.position.lat, params.position.lon, params.position.alt)
+  
+  local temperature, pressure = atmosphere.getTemperatureAndPressure(point)
+
+  return GRPC.success(
+    {
+      temperature = temperature, -- Kelvin
+      pressure = pressure -- Pascals
+    }
+  )
+end
--- a/README.md
+++ b/README.md
@@ -54,13 +54,13 @@ dofile(GRPC.basePath .. [[grpc.lua]])
 Test the running server via [grpcurl](https://github.com/fullstorydev/grpcurl): (Remove the `.exe` when running on Linux)
 
 ```bash
-grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs_mission.proto -d '{\"text\": \"Works!\", \"display_time\": 10, \"clear_view\": false}' 127.0.0.1:50051 dcs.Triggers/OutText
+grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"text\": \"Works!\", \"display_time\": 10, \"clear_view\": false}' 127.0.0.1:50051 dcs.Triggers/OutText
 ```
 
 or watch the mission event stream via:
 
 ```bash
-grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs_mission.proto -d '{}' 127.0.0.1:50051 dcs.Mission/StreamEvents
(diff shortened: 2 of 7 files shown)


lua/methods/atmosphere.lua
```diff
@@ -0,0 +1,56 @@
...
+  
+  if direction < 0 then
+    direction = direction + 360
+  end
+  
+  -- Convert TO direction to FROM direction. 
+  if direction > 180 then
+    direction = direction-180
+  else
+    direction = direction+180
```
rkusa, line 21: This will convert `180` to `360` and not `0`, which is fine for me, just want to mention it in case you were aiming at a `000` rather than `360`

lua/methods/atmosphere.lua
```diff
@@ -0,0 +1,56 @@
...
+  end
+  
+  -- Calc 2D strength.
+  local strength = math.sqrt((windvec3.x)^2+(windvec3.z)^2)
+
+  return  {
+    heading = direction, -- Western style. Heading the wind is coming FROM
+    strength = strength -- m/s
```
rkusa, line 29: Nit: how about `speed` as a name?

lua/methods/atmosphere.lua
```diff
@@ -0,0 +1,56 @@
...
+    direction = direction-180
+  else
+    direction = direction+180
+  end
+  
+  -- Calc 2D strength.
+  local strength = math.sqrt((windvec3.x)^2+(windvec3.z)^2)
+
+  return  {
+    heading = direction, -- Western style. Heading the wind is coming FROM
```
rkusa, line 28: I think `direction` is usually used to denote the direction the wind is coming from, while `heading` would the direction the wind is going to

rkusa (approved):
Only added a few minor comments 👍

> * The DCS provided values for temperature and pressure (Kelvin and Pascals)

- As a client-author/-user would you see any benefit if we'd let the user request a specific unit and the server does the conversion?
- I am wondering if we should bake the unit somehow into the response (regardless of the above point), ala: `{"pressure_in_pascal": 42}` or `{"pressure": {"value":42,"unit":"pascal"}}`. I am not sure about how I really feel about this myself yet, do you have an opinion?

lua/methods/atmosphere.lua
rurounijones: Good catch. This is intentional.

rurounijones:
> 
> 
> Only added a few minor comments 👍
> 
> > * The DCS provided values for temperature and pressure (Kelvin and Pascals)
> 
>     * As a client-author/-user would you see any benefit if we'd let the user request a specific unit and the server does the conversion?
> 
>     * I am wondering if we should bake the unit somehow into the response (regardless of the above point), ala: `{"pressure_in_pascal": 42}` or `{"pressure": {"value":42,"unit":"pascal"}}`. I am not sure about how I really feel about this myself yet, do you have an opinion?

I am really against this. Adding more code and complexity to the lua is something I think we should avoid at all costs:
* It complicates stuff that has to happen in the mission environment which I think we should strive to avoid generally as a rule.
* It needs to be written in lua which, frankly, I detest.

Let us stick to returning SI units and let the clients convert them to their preferred units in their own, much nicer languages :grin:.

lua/methods/atmosphere.lua
rurounijones: I wanted to avoid direction because that is what was used in the ME but yeah we should probably stick to it. Changed "strength" to "speed"

lua/methods/atmosphere.lua
rurounijones: Done

rkusa:
> I am really against this. Adding more code and complexity to the lua is something I think we should avoid at all costs:
> 
> * It complicates stuff that has to happen in the mission environment which I think we should strive to avoid generally as a rule.
> * It needs to be written in lua which, frankly, I detest.
> 
> Let us stick to returning SI units and let the clients convert them to their preferred units in their own, much nicer languages 😁.

I would definitely avoid doing this in Lua, too. But we could also do it in Rust, if we want to have the feature. I personally don't need the feature, but it would be easy to add which is why I brought it up.

rurounijones:
> 
> 
> > I am really against this. Adding more code and complexity to the lua is something I think we should avoid at all costs:
> > 
> > * It complicates stuff that has to happen in the mission environment which I think we should strive to avoid generally as a rule.
> > * It needs to be written in lua which, frankly, I detest.
> > 
> > Let us stick to returning SI units and let the clients convert them to their preferred units in their own, much nicer languages 😁.
> 
> I would definitely avoid doing this in Lua, too. But we could also do it in Rust, if we want to have the feature. I personally don't need the feature, but it would be easy to add which is why I brought it up.

Ah right, ok. I think we should stick to just returning SI units otherwise we open ourselves up to the possibility of requests for other things to be returned in different units (e.g. altitudes returned in feet, distances in nautical miles).

If we stick to a clear "SI only" policy we can avoid having to entertain those requests.

rurounijones:
Only just noticed that this was approved so merging despite the open questions since I believe we can come back to those later if needed (and would require work by @rkusa if it is implemented in Rust, anyway).

PR #14, opened 2021-07-23
Add MarkPoint management APIs

rurounijones:
Add APIs that allows for managing MarkPoints in the mission environment.

Tested using grpcurl. Examples below

```plain
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"id\": 14001, \"text\": \"Scripted markpoint\", \"position\": { \"lat\": 43.116012753760046, \"lon\": 41.96052007622479, \"alt\": 6089.67431640625 }, \"coalition\": 2, \"readOnly\": false, \"message\": \"\"}' 127.0.0.1:50051 dcs.Triggers/MarkToCoalition
{}
```

```plain
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"id\": 14001}' 127.0.0.1:50051 dcs.Triggers/RemoveMark
{}
```

### DIFF
--- a/lua/methods/trigger.lua
+++ b/lua/methods/trigger.lua
@@ -19,3 +19,33 @@ GRPC.methods.setUserFlag = function(params)
   trigger.action.setUserFlag(params.flag, params.value)
   return GRPC.success(nil)
 end
+
+GRPC.methods.markToAll = function(params)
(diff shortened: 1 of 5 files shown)


rkusa (approved):
Haven't tested it, but code looks good - having a reusable `EmptyResponse` is a good idea 👍

PR #16, opened 2021-07-24
Add World.getMarkPanels API

rurounijones:
Add the World.getMarkPanels API so that we can implemented all the
MarkPanel APIs.

I have only tested in single player where it appears that the initiator
is _always_ the player even if the MarkPoint was created via scripting.

It also looks like, in SP at least, that "MarkToAll" API will result in
the markpoint being the "Blue" coalition.

I haven't yet had the opportunity to test on an MP server since it is not as easy as SP. Not sure if you prefer to hold off on merging until that is done. 
 
 ```plain
./grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{}' 127.0.0.1:50051 dcs.World/GetMarkPanels
{
	"markPanels": [{
		"id": 1301,
		"time": 28916,
		"initiator": {
			"id": 1,
			"name": "Aerial-1-1",
			"callsign": "Enfield11",
			"coalition": "BLUE",
			"type": "FA-18C_hornet",
			"position": {
				"lat": 42.18262219221839,
				"lon": 42.476887902005345,
				"alt": 46.84111404418945
			},
			"playerName": "New callsign"
		},
		"coalition": "RED",
		"groupID": 4294967295,
		"text": "Scripted markpoint",
		"position": {
			"lat": 43.116012753760046,
			"lon": 41.96052007622479,
			"alt": 2683.032958984375
		}
	}, {
		"id": 13000,
		"time": 28835,
		"initiator": {
			"id": 1,
			"name": "Aerial-1-1",
			"callsign": "Enfield11",
			"coalition": "BLUE",
			"type": "FA-18C_hornet",
			"position": {
				"lat": 42.18262219221839,
				"lon": 42.476887902005345,
				"alt": 46.84111404418945
			},
			"playerName": "New callsign"
		},
		"groupID": 4294967295,
		"text": "Scripted markpoint",
		"position": {
			"lat": 43.116012753760046,
			"lon": 41.96052007622479,
			"alt": 2683.032958984375
		}
	}]
}
```

### DIFF
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -69,3 +69,23 @@ end
 GRPC.exporters.object = function(object)
   return {}
 end
+
+GRPC.exporters.markPanel = function(markPanel)
+  local mp = {
+    id = markPanel.idx, 
+    time = markPanel.time,
+    initiator = GRPC.exporters.unit(markPanel.initiator),
+    text = markPanel.text,
+    position = toLatLonPosition(markPanel.pos)
+  }
+
+  if (markPanel.coalition >= 0 and markPanel.coalition <= 2) then
+  	mp["coalition"] = markPanel.coalition;
+  end
+
+  if (markPanel.groupID > 0) then
+  	mp["groupId"] = markPanel.groupID;
+  end
+
+  return mp
+end
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -8,15 +8,6 @@
 local GRPC = GRPC
 local coord = coord
 
-local function toLatLonPosition(pos)
-  local lat, lon, alt = coord.LOtoLL(pos)
-  return {
-    lat = lat,
-    lon = lon,
-    alt = alt,
-  }
-end
-
 GRPC.exporters.unit = function(unit)
   return {
     id = tonumber(unit:getID()),
@@ -24,7 +15,7 @@ GRPC.exporters.unit = function(unit)
     callsign = unit:getCallsign(),
     coalition = unit:getCoalition(),
     type = unit:getTypeName(),
-    position = toLatLonPosition(unit:getPoint()),
+    position = GRPC.toLatLonPosition(unit:getPoint()),
     playerName = unit:getPlayerName()
   }
 end
@@ -33,7 +24,7 @@ GRPC.exporters.weapon = function(weapon)
   return {
     id = tonumber(weapon:getName()),
     type = weapon:getTypeName(),
-    position = toLatLonPosition(weapon:getPoint()),
+    position = GRPC.toLatLonPosition(weapon:getPoint()),
   }
 end
 
@@ -48,7 +39,7 @@ GRPC.exporters.airbase = function(airbase)
     coalition = airbase:getCoalition(),
     category = airbase:getDesc()['category'],
     displayName = airbase:getDesc()['displayName'],
-    position = toLatLonPosition(airbase:getPoint())
+    position = GRPC.toLatLonPosition(airbase:getPoint())
   }
 
   if airbase:getUnit() then
@@ -76,7 +67,7 @@ GRPC.exporters.markPanel = function(markPanel)
     time = markPanel.time,
     initiator = GRPC.exporters.unit(markPanel.initiator),
     text = markPanel.text,
-    position = toLatLonPosition(markPanel.pos)
+    position = GRPC.toLatLonPosition(markPanel.pos)
   }
 
   if (markPanel.coalition >= 0 and markPanel.coalition <= 2) then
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: protos/world.proto [hunk 1]
```diff
@@ -10,4 +10,10 @@ message GetAirbasesRequest {
 
 message GetAirbasesResponse {
   repeated Airbase airbases = 1;
+}
+
+message GetMarkPanelsRequest {}
+
+message GetMarkPanelsResponse {
+	repeated MarkPanel markPanels = 1;
```
rkusa, line 18:
```suggestion
	repeated MarkPanel mark_panels = 1;
```

protos/common.proto
```diff
@@ -82,4 +82,14 @@ message Target {
...
-// End of Object subtypes
\ No newline at end of file
+// End of Object subtypes
+
+message MarkPanel {
+	uint32 id = 1; 
+	double time = 2; // I have no idea what this time value means
+	Unit initiator = 3;
+	optional Coalition coalition = 4;
+	optional uint32 groupID = 5;
```
rkusa, line 92:
```suggestion
	optional uint32 group_id = 5;
```

lua/exporters/object.lua
```diff
@@ -68,4 +68,24 @@ end
...
   return {}
+end
+
+GRPC.exporters.markPanel = function(markPanel)
+  local mp = {
+	id = markPanel.idx, 
+	time = markPanel.time,
+    initiator = GRPC.exporters.unit(markPanel.initiator),
+	text = markPanel.text,
+	position = toLatLonPosition(markPanel.pos)
```
rkusa, line 79:
```suggestion
    id = markPanel.idx, 
    time = markPanel.time,
    initiator = GRPC.exporters.unit(markPanel.initiator),
    text = markPanel.text,
    position = toLatLonPosition(markPanel.pos)
```

protos/world.proto
```diff
@@ -10,4 +10,10 @@ message GetAirbasesRequest {
 
 message GetAirbasesResponse {
   repeated Airbase airbases = 1;
+}
+
+message GetMarkPanelsRequest {}
```
rkusa, line 15: Would it make sense to rename `EmptyResponse` to `Empty` and re-use it for empty responses too?

rkusa (commented):
Just some minor casing/indentation comments 👍

protos/world.proto
rurounijones: My C# muscle memory is beating out by Ruby muscle memory. Oh no...

Changed.

protos/common.proto
rurounijones: Fixed

protos/world.proto
rurounijones: So I have been reading gRPC best practices as created by teams that have been using gRPC for years and their recommendations can be summed up as `SHARE NOTHING".

Their recommendation is that if you have a Service with a method called `GetMarkPanels` then your request and response will be exactly that name with the `Request` and `Response` suffixes.

This greatly improves backwards compatibility and autogeneration / documentation possibilities. While I don't think this is a huge benefit for _us_ since we are unlikely to be changing these APIs I am loath to go against these recommendations.

Due to this I am thinking that after this commit and your Unit streaming commit I will need to go through everything and changed the requests and responses if we want to follow the recommendations.

What do you think?

* https://medium.com/@akhaku/protobuf-definition-best-practices-87f281576f31
* https://docs.buf.build/style-guide/
* https://docs.buf.build/lint-rules#rpc_request_standard_name-rpc_response_standard_name-rpc_request_response_unique

rurounijones:
Squashed commits. Thanks for making the suggested changes to cut down time.

protos/world.proto
rkusa: I am totally up for following best practices, thanks for going through them 👍

lua/exporters/object.lua
```diff
@@ -68,4 +68,24 @@ end
...
\ No newline at end of file
+end
+
+GRPC.exporters.markPanel = function(markPanel)
+  local mp = {
+    id = markPanel.idx, 
+    time = markPanel.time,
+    initiator = GRPC.exporters.unit(markPanel.initiator),
+    text = markPanel.text,
+    position = toLatLonPosition(markPanel.pos)
```
rkusa, line 79: Just a minor thing, because I'll probably not remember the line when merging my PR in some weeks. Would you mind changing this to the following already?

```suggestion
    position = GRPC.toLatLonPosition(markPanel.pos)
```

rkusa (approved):
👍

lua/exporters/object.lua
rurounijones: Done

PR #18, opened 2021-07-25
Add more trigger APIs

rurounijones:
Add OutText commands for coalitions and groups. I have not added
"ForCountry" because I have never seen it used and it requires creating
a country Enum which is a lot of tedious typing.

Also add the APIs for the pretty visual triggers: Signal Flares, Smoke,
Illumination Bombs and EXPLOSIONS.

Tested in-game (and it was very pretty)

### DIFF
--- a/lua/methods/trigger.lua
+++ b/lua/methods/trigger.lua
@@ -9,6 +9,18 @@ GRPC.methods.outText = function(params)
   return GRPC.success(nil)
 end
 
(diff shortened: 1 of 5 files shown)


rkusa (approved):
> I have not added
> "ForCountry" because I have never seen it used and it requires creating
> a country Enum which is a lot of tedious typing.

True, never seen in either, so I think it is perfectly reasonable to leave it out 👍

PR #22, opened 2021-08-26
implement repl

rkusa:
A REPL into the DCS scripting environment is something I am using quite heavily during development, so I gave it a try to implement it for the gRPC server.

image

Usage:

```bash
cargo build --bin repl
# start DCS mission
cargo run --bin repl
```

Wdyt, should we keep it?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,10 @@ version = "0.1.0"
 authors = ["Markus Ast <m@rkusa.st>"]
 license = "AGPL-3.0-or-later"
 edition = "2018"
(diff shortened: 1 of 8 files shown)


rurounijones:
> A return by default would be nice, but I am not sure if we can reliably place it correctly when someone executes a multi-line lua expression. We could try the following:
>    if single line, place a return if it is not already there
>    if multi-line, don't do anything (or place it for the last line, not sure)

Good points. Lets Keep It Simple

rurounijones (approved):

PR #28, opened 2021-09-04
add missing events and event properties

rkusa:
I went through all events, testing them, adding missing ones, and adding missing properties. I've ignored all events I couldn't find a way to trigger.

### DIFF
--- a/lua/methods/mission.lua
+++ b/lua/methods/mission.lua
@@ -5,17 +5,18 @@ local function exporter(object)
 
   local category = object:getCategory()
 
-  if(category == Object.Category.UNIT) then
-    return GRPC.exporters.unit(object)
-  elseif(category == Object.Category.WEAPON) then
-    return GRPC.exporters.weapon(object)
-  elseif(category == Object.Category.STATIC) then
-    return GRPC.exporters.static(object)
-  elseif(category == Object.Category.BASE) then
+  if category == Object.Category.BASE or object.className_ == 'Airbase' then
+    -- carriers are of category unit, but are a Airbase class
     return GRPC.exporters.airbase(object)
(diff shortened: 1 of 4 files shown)


lua/methods/mission.lua
```diff
@@ -5,17 +5,18 @@ local function exporter(object)
...
-  elseif(category == Object.Category.STATIC) then
+  elseif category == Object.Category.STATIC then
     return GRPC.exporters.static(object)
-  elseif(category == Object.Category.BASE) then
-    return GRPC.exporters.airbase(object)
-  elseif(category == Object.Category.SCENERY) then
+  elseif category == Object.Category.SCENERY then
     return GRPC.exporters.scenery(object)
-  elseif(category == Object.Category.Cargo) then
+  elseif category == Object.Category.Cargo then
```
rkusa, line 19: Removed the parenthesis for consistency with the other Lua code. But I'd also be fine to change everything to using parenthesis instead

rurounijones: Parethessislessness is fine :grin:

src/stream.rs
```diff
@@ -133,6 +133,7 @@ async fn handle_event(state: &mut State, event: Event) -> Result<(), Error> {
                 Some(Initiator {
                     initiator: Some(initiator::Initiator::Unit(unit)),
                 }),
+            ..
```
rurounijones, line 136: Is this supposed to be here?

rkusa: Yes, the struct has now another property, something the compiler would complain about. This is the syntax for telling the compiler that we don't care about any other property of the struct 👍

rurounijones: Nice, Today I learned.

PR #29, opened 2021-09-05
add stacktraces to event and req handler errors

rkusa:
This change adds stacktraces to event and request handlers. This makes it way easier to pin-point the cause of errors.

Do you have any concerns with the usage of `xpcall` or the `debug` module? Feels a bit like a waste to create a new inline function each time `xpcall` is used, but I couldn't find another way to add stacktraces in Lua 5.1.

### DIFF
--- a/lua/grpc.lua
+++ b/lua/grpc.lua
@@ -115,7 +115,7 @@ local function handleRequest(method, params)
   local fn = GRPC.methods[method]
 
   if type(fn) == "function" then
-    local ok, result = pcall(fn, params)
+    local ok, result = xpcall(function() return fn(params) end, debug.traceback)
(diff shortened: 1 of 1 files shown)


rurounijones:
> Do you have any concerns with the usage of `xpcall` or the `debug` module? Feels a bit like a waste to create a new inline function each time `xpcall` is used, but I couldn't find another way to add stacktraces in Lua 5.1.

Honestly I do not know enough about Lua to say one way or the other. I agree it _sounds_ a bit wierd but the benefits are really handy and I have no idea if there is a performance impact or not.

It is a simple commit that can always be reverted in the future if needed so I say merge it.

Issue #19, opened 2021-08-21
DCS Server crashes on player connection


rurounijones:
The DCS dedicated server will crash with the [current](https://github.com/DCS-gRPC/rust-server/commit/4e6e472dcdc68cfca4c20fbafd4f342a1ccaf6ac) DCS-gRPC master branch code on player connection:

Steps to replicate.
1. Run a mission, with DCS-gRPC enabled, on a DCS **dedicated server**.
2. Connect to the running server in DCS.
3. Server will instantly crash with the below exception.

Note that this does not happen:
* when running the mission in singleplayer so it must be something to do with a multiplayer client connecting.
* when no one connects, the dedicated server will happily run, even with complicated AI actions going on, as long as no one connects.

I am not sure when this started happening, there is at least one version of the server running on the Hoggit Syria At War server with no issues so we know of so [this](https://github.com/DCS-gRPC/rust-server/tree/7fd835d5cddbd7df07eaf6c1e86d5768d6d72850) commit is working correctly.

If you cannot think of anything obvious then I wil do a `git bisect` although that will take a bit of time.

Stack trace below:

```
2021-08-21 02:27:34.651 INFO    EDCORE: signal caught: 'Abnormal termination'
2021-08-21 02:27:34.651 INFO    EDCORE: try to write dump information
2021-08-21 02:27:34.653 INFO    EDCORE: # -------------- 20210821-022734 --------------
2021-08-21 02:27:34.654 INFO    EDCORE: DCS/2.7.5.10869 (x86_64; Windows NT 10.0.19042)
2021-08-21 02:27:34.656 INFO    EDCORE: C:\Windows\System32\KERNELBASE.dll
2021-08-21 02:27:34.657 INFO    EDCORE: # 00000000 STATUS_SUCCESS at 73F44ED9 00:00000000
2021-08-21 02:27:34.663 INFO    EDCORE: SymInit: Symbol-SearchPath: '.;C:\Program Files\Eagle Dynamics\DCS World OpenBeta Server;C:\Program Files\Eagle Dynamics\DCS World OpenBeta Server\bin;C:\Windows;C:\Windows\system32;SRV*C:\websymbols*https://msdl.microsoft.com/download/symbols;', symOptions: 528, UserName: 'CowServer'
2021-08-21 02:27:34.665 INFO    EDCORE: OS-Version: 10.0.19042 () 0x100-0x1
[...]

PR #30, opened 2021-09-05
log everything into gRPC.log

rkusa:
I was a bit annoyed that I always have to watch both `dcs.log` and `gRPC.log` to find all error/warning logs. This PR makes sure that everything is logged into `gRPC.log`.

Wdyt?

### DIFF
--- a/lua/grpc.lua
+++ b/lua/grpc.lua
@@ -44,6 +44,20 @@ GRPC.error = function(msg)
   }
 end
 
+--
+-- Logging methods
+--
+
+GRPC.logError = function(err)
+  grpc.log_error(err)
+  env.error("[GRPC] "..err)
+end
+
+GRPC.logWarning = function(err)
+  grpc.log_warning(err)
+  env.warning("[GRPC] "..err)
+end
+
 --- The client specified an invalid argument
 GRPC.errorInvalidArgument = function(msg)
   return {
@@ -119,7 +133,7 @@ local function handleRequest(method, params)
     if ok then
       return result
     else
-      env.error("[GRPC] error executing "..method..": "..tostring(result))
+      GRPC.logError("error executing "..method..": "..tostring(result))
       return {
         error = tostring(result)
       }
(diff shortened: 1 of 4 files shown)


lua/grpc.lua
```diff
@@ -44,6 +44,20 @@ GRPC.error = function(msg)
...
+--
+
+GRPC.logError = function(err)
+  grpc.log_error(err)
+  env.error("[GRPC] "..err)
+end
+
+GRPC.logWarning = function(err)
+  grpc.log_warning(err)
+  env.warning("[GRPC] "..err)
```
rkusa, line 58: Logs it into both `dcs.log` and `gRPC.log`. I'd also be fine though to remove the logs from `dcs.log`.

rurounijones: I think having warnings and exceptions in the main dsc.log reduces work for server admins. Everything else being in grpc.log is fine I think.

rurounijones (approved):

lua/grpc.lua
rurounijones: Oh, it just occurred to me. Can we keep the GRPC startup message appearing in DCS.log?

rkusa: Yes the startup messages (`loading ...` and `loaded ...`) are kept in the `dcs.log` - I also think that they are very important to even know if the gRPC module started or not, as we cannot even expect logs in the `gRPC.log` file otherwise.

rkusa: > I think having warnings and exceptions in the main dsc.log reduces work for server admins. Everything else being in grpc.log is fine I think.

Most errors and warnings would be mirrored into both files, but there are already errors that will only make it into the `gRPC.log` file. They are the ones that are logged on the Rust side. One of them being the fix for https://github.com/DCS-gRPC/rust-server/issues/19 as an example.

rkusa: My plan is btw to hook up the logging to https://sentry.io next, as I rarely skim logs for errors unless I've explicitly noticed something not working.

rkusa:

lua/grpc.lua
rurounijones: Could we avoid hooking directly up to certain sinks?

If we can make sure that the server outputs JSON logs to file then another separate application can then ship those json logs to the sinks of the admins choice. I know that Hoggit uses an ELK stack for example.

PR #33, opened 2021-09-06
extend gRPC server's access to DCS' hook env

rkusa:
I am have to admit, I am still surprised that this ended up working, which makes be a bit suspicious. I am thus considering this a draft for now, and am planning on doing some more stability tests and some reading on certain things I don't quite understand why they work.

This PR:
- adds a DCS hook (lua file into `Scripts/Hooks/`) which allows handling RPC calls inside of the hooks environment (does not need a second server)
- implements hot-reloading of the DLL to still be able to recompile the DLL between mission restarts (I'd hate having to restart DCS for each change)
- adds `dcs.Hook/GetMissionName` as a first example for a method calling into the hook

More details will follow as inline comments.

Example to test that both envs still work:

```
.\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"text\": \"Works!\", \"display_time\": 10, \"clear_view\": false}' 127.0.0.1:50051 dcs.Triggers/OutText

.\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{}' 127.0.0.1:50051 dcs.Hook/GetMissionName
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,14 @@ build: export LUA_LIB_NAME=lua
 build: export LUA_LIB=$(CURDIR)/lua/lua5.1/
 build: export LUA_INC=$(CURDIR)/lua/lua5.1/include
 build:
-	cargo build
+	cargo build --features hot-reload
+	powershell copy target/debug/dcs_grpc_server.dll target/debug/dcs_grpc_server_hot_reload.dll
 
 watch: export LUA_LIB_NAME=lua
 watch: export LUA_LIB=$(CURDIR)/lua/lua5.1/
 watch: export LUA_INC=$(CURDIR)/lua/lua5.1/include
 watch:
-	cargo watch
+	cargo watch -x "check --features hot-reload"
 
 test: export LUA_LIB_NAME=lua
 test: export LUA_LIB=$(CURDIR)/lua/lua5.1/
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ mission on a DCS server.
 
 Download the latest version of the server from the [Releases](https://github.com/DCS-gRPC/rust-server/releases) and
 extract the zip file into your DCS Server directory. This is typically found in
-`C:\Users\USERNAME\Saved Games\DCS.openbeta_server`. Once extracted you will have a `Scripts\DCS-gRPC` folder and
-a `Mods\Tech\DCS-gRPC` folder.
+`C:\Users\USERNAME\Saved Games\DCS.openbeta_server`. Once extracted you will have a `Scripts\DCS-gRPC` folder,
+a `Mods\Tech\DCS-gRPC` folder, and a `Scripts\Hooks\DCS-gRPC.lua` file.
 
 ### MissionScripting Sanitisation Removal
 
@@ -34,22 +34,7 @@ end
 
 ### Mission Editing
 
-Add the following code to your mission. This will start the DCS-gRPC server. You can add this code to a DOSCRIPT
-trigger in the mission editor or include it in your own scripts if you already have some.
-
-```lua
-package.cpath = package.cpath..lfs.writedir()..[[Mods\tech\DCS-gRPC\?.dll;]]
-GRPC = { basePath = lfs.writedir()..[[Scripts\DCS-gRPC\]] }
-
-local luaPath = GRPC.basePath .. [[grpc.lua]]
-local f = assert( loadfile(luaPath) )
-
-if f == nil then
-  error ("[GRPC]: Could not load " .. luaPath )
-else
-  f()
-end
-```
+Create a trigger of type `MISSION START` and add a `DO SCRIPT FILE` action loading `Scripts\DCS-gRPC\grpc-mission.lua`.
 
 ### Confirmation
 
@@ -82,7 +67,9 @@ $env:LUA_INC=(Get-Item -Path ".\").FullName+"/lua/lua5.1/include"
 cargo build
 ```
 
-### MissionScripting Sanitisation Removal
+### Link files
+
+To always have the latest development changes
 
 DCS-gRPC requires the removal of sanitisation features in DCS scripting.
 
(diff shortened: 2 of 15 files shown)


Makefile
```diff
@@ -2,13 +2,14 @@ build: export LUA_LIB_NAME=lua
 build: export LUA_LIB=$(CURDIR)/lua/lua5.1/
 build: export LUA_INC=$(CURDIR)/lua/lua5.1/include
 build:
-	cargo build
+	cargo build --features hot-reload
+	powershell copy target/debug/dcs_grpc_server.dll target/debug/dcs_grpc_server_hot_reload.dll
```
rkusa, line 6: `dcs_grpc_server.dll` and `target/debug/dcs_grpc_server_hot_reload.dll` will be the same with the difference that DCS loads and locks `dcs_grpc_server_hot_reload.dll`, while `dcs_grpc_server.dll` in return loads `dcs_grpc_server_hot_reload.dll`. If a mission is stopped, `dcs_grpc_server.dll` release its lock on `dcs_grpc_server_hot_reload.dll`, which in the end allows to re-compile the code between mission restarts instead of having to restart DCS>

README.md
```diff
@@ -34,22 +34,7 @@ end
...
-local luaPath = GRPC.basePath .. [[grpc.lua]]
-local f = assert( loadfile(luaPath) )
-
-if f == nil then
-  error ("[GRPC]: Could not load " .. luaPath )
-else
-  f()
-end
-```
+Create a trigger of type `MISSION START` and add a `DO SCRIPT FILE` action loading `Scripts\DCS-gRPC\grpc-mission.lua`.
```
rkusa, line 37: I thought it might be easier to just load a Lua file, wdyt? I'd also be fine with changing it back.

README.md
```diff
@@ -102,16 +89,30 @@ do
...
+```
+
+Create directories and links:
+
+```ps1
+New-Item -ItemType SymbolicLink -Path "M:\Saved Games\DCS.openbeta\Scripts\DCS-gRPC" -Value "M:\Development\DCS-gRPC\rust-server\lua"
+New-Item -ItemType SymbolicLink -Path "M:\Saved Games\DCS.openbeta\Scripts\Hooks\DCS-gRPC.lua" -Value "M:\Development\DCS-gRPC\rust-server\lua\grpc-hook.lua"
+New-Item -Path "M:\Saved Games\DCS.openbeta\Mods\Tech\DCS-gRPC" -ItemType "directory"
+New-Item -ItemType SymbolicLink -Path "M:\Saved Games\DCS.openbeta\Mods\Tech\DCS-gRPC\dcs_grpc_server.dll" -Value "M:\Development\DCS-gRPC\rust-server\target\debug\dcs_grpc_server.dll"
+New-Item -ItemType SymbolicLink -Path "M:\Saved Games\DCS.openbeta\Mods\Tech\DCS-gRPC\dcs_grpc_server_hot_reload.dll" -Value "M:\Development\DCS-gRPC\rust-server\target\debug\dcs_grpc_server_hot_reload.dll"
```
rkusa, line 109: Todo:
- [ ] use paths that are not specific to my `M:` windows installation

Wdyt about the symbolic links instead of the difference Lua mission code?

src/lib.rs
```diff
@@ -24,7 +26,8 @@ static INITIALIZED: Lazy<AtomicBool> = Lazy::new(|| AtomicBool::new(false));
 static SERVER: Lazy<RwLock<Option<Server>>> = Lazy::new(|| RwLock::new(None));
 
 struct Server {
-    ipc: IPC<Event>,
+    ipc_mission: IPC<Event>,
+    ipc_hook: IPC<()>,
```
rkusa, line 30: We now basically have two channels. One for the mission and one for the hook. This is how we don't need two actual TCP servers.

README.md
rurounijones: We need something that is fail-safe. Not everyone who works on a mission will have DCS-gRPC installed locally. (e.g. people modifying and testing a Hoggit mission where gRPC export is not needed so we cannot assume that `grpc-mission.lua` is present.

I think for maximum nicety to mission devs I would go with the above `DO SCRIPT` instead of `DO SCRIPT FILE`

README.md
rurounijones: I would update the README to just include both options and let people choose. Personally I would not use the Symbolic links since I want some missions to use my "dev" version and some missions to use the latest "Stable" version from main so I edit the lua in each of the missions to point to different locations..

src/lib.rs
rurounijones: That is exceedintly cool. Is the limit of calling into the mission env then shared between the two?

Makefile
rurounijones: I do my testing in a local single-player game of DCS and as long as the mission is not running I don't actually have this issue as far as I remember.

However having the extra ability is a great idea.

rkusa: > I do my testing in a local single-player game of DCS and as long as the mission is not running I don't actually have this issue as far as I remember.

Me too. Previously, unloading the mission also unlocked the DLL. But once the hook env loads the DLL, it locks it until DCS is closed. So we didn't have the issue previously, but the hook addition would have introduced it if it weren't for the hot reloading.

src/lib.rs
rkusa: Each env comes with its own limits for now. The main challenge for a shared limit would probably to figure out a way of distribution the calls in a fair manner between both (and to not create dead-locks if client calls to the mission env depend on a call to the hook env)

Makefile
rurounijones: Aaaah right, gotcha.

rkusa:

rkusa:
> I [...] am planning on doing some more stability tests and some reading on certain things I don't quite understand why they work.

I read up on the memory of DLL's and all makes sense know - i.e., it works like it is supposed to work. I also couldn't find any stability issues, but I wouldn't be surprised if some might come up in the future - so I'll keep an eye on it.

README.md
rkusa: Kept both 👍

README.md
rkusa: Kept both 👍

rurounijones (approved):
Noice, whole new load of possibilities!

rkusa:

PR #35, opened 2021-09-12
implement chat filter POC

rkusa:
This PR adds a POC for filtering chat messages using the new access to the hook env.

An example client that converts all messages to uppercase can be run via `cargo run --example chat_filter`.

We might not wanna keep the POC though, see inline comments.

### DIFF
--- a/examples/chat_filter.rs
+++ b/examples/chat_filter.rs
@@ -1,0 +1,71 @@
+use std::time::Duration;
+
+use backoff::ExponentialBackoff;
+use dcs_grpc_server::rpc::dcs::hook::{FilterChatMessageRequest, StreamChatRequest};
+use dcs_grpc_server::rpc::dcs::hook_client::HookClient;
+use futures_util::future::{select, FutureExt};
+use futures_util::stream::StreamExt;
+use log::LevelFilter;
+use tonic::{transport, Request, Status};
+
+async fn run() -> Result<(), Error> {
+    let addr = "http://127.0.0.1:50051";
+    log::debug!("Connecting to gRPC server at {}", addr);
+
+    let endpoint = transport::Endpoint::from_static(addr).keep_alive_while_idle(true);
+    let mut client = HookClient::connect(endpoint).await?;
+    let mut chat_messages = client
+        .stream_chat(Request::new(StreamChatRequest {}))
+        .await?
+        .into_inner();
+
+    while let Some(msg) = chat_messages.next().await {
+        let msg = msg?;
+        client
+            .filter_chat_message(Request::new(FilterChatMessageRequest {
+                uuid: msg.uuid,
+                message: msg.message.to_uppercase(),
+            }))
+            .await?;
+    }
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() {
+    pretty_env_logger::formatted_builder()
+        .filter_module("chat_filter", LevelFilter::Debug)
+        .init();
+
+    let backoff = ExponentialBackoff {
+        // never wait longer than 30s for a retry
+        max_interval: Duration::from_secs(30),
+        // never stop trying
+        max_elapsed_time: None,
+        ..Default::default()
+    };
+
+    select(
+        Box::pin(backoff::future::retry_notify(
+            backoff,
+            || async { run().await.map_err(backoff::Error::Transient) },
+            |err, backoff: Duration| {
+                log::error!("Error: {}", err);
+                log::info!("Retrying after error in {:.2}s", backoff.as_secs_f64());
+            },
+        )),
+        Box::pin(tokio::signal::ctrl_c().map(|_| ())),
+    )
+    .await;
+}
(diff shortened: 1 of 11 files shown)


lua/grpc-hook.lua
```diff
@@ -33,6 +33,10 @@ local function init()
     _G.GRPC = nil
   end
 
+  function handler.onPlayerTrySendChat(playerID, msg, all)
```
rkusa, line 36: The docs about `onPlayerTrySendChat` say:

> they are called directly from the network code, so try to make them as fast as possible

And this is unfortunately true. The handler blocks the server until `onPlayerTrySendChat` returns. This is why we might not wanna keep the POC.

src/chat.rs
```diff
@@ -0,0 +1,79 @@
...
+            player_id,
+            message: message.clone(),
+            all,
+        };
+        if self.stream.send(chat_message).is_err() {
+            // all receivers are dropped, return message unfiltered
+            return message;
+        }
+
+        // wait for up to 16 milliseconds for a filter request from the client
```
rkusa, line 63: To prevent the server from getting blocked too long, the client only has 16ms to send a filter request. 16ms can be fine tuned. I went with 16ms because it corresponds to one whole frame if the server is running at 60fps (`1000/16=~16`).

examples/chat_filter.rs
```diff
@@ -0,0 +1,71 @@
...
+        .into_inner();
+
+    while let Some(msg) = chat_messages.next().await {
+        let msg = msg?;
+        client
+            .filter_chat_message(Request::new(FilterChatMessageRequest {
+                uuid: msg.uuid,
+                message: msg.message.to_uppercase(),
+            }))
+            .await?;
```
rkusa, line 29: This best shows how the implementation works. The same approach could be used to implement a slot blocking, or banning of users.

1. The client streams all chat messages, where each chat message has a UUID
2. The server expects the client to send a `FilterChatMessageRequest ` within 16ms that tells the server how to proceed with the message

rurounijones:
Is this a POC that is applicable to many use-cases? In this case it is a server-initiated round trip from the server to the network client and back to the server with the time-limit.

I don't know much about the hook env APIs but this seems like it would be an edge-case? My image is that most of the hook env APIs would be things like streaming data or receiving an incoming request. No round trips involved.

rkusa:
The POC is applicable to only the following APIs - those are the only ones that need a time-limited round trip:

```lua
function onPlayerTryConnect(addr, name, ucid, playerID) --> true | false, "disconnect reason"
    return true
end

function onPlayerTrySendChat(playerID, msg, all) -- -> filteredMessage | "" - empty string drops the message
    return msg
end

function onPlayerTryChangeSlot(playerID, side, slotID) -- -> true | false
    return true
end
```

Possible use-cases for those APIs I can think of that require a time-limited round trip:
- `onPlayerTryConnect`:
  - ban users
  - VIP slots
- `onPlayerTrySendChat`:
  - filter chat messages
  - modify chat message
  - receive commands, where you make sure that they are received but are not listed in the chat history
- `onPlayerTryChangeSlot`:
  - slot block

> I don't know much about the hook env APIs but this seems like it would be an edge-case? My image is that most of the hook env APIs would be things like streaming data or receiving an incoming request. No time-limited round trips involved.

Everything else besides the things listed above, are the same as for the mission env, i.e. as you said: "streaming data or receiving an incoming request".

> (I am not sure if the slot-blocking has the same time-limit block-the-server approach for example. Could we instead receive the slot occupied message then remove the user from it at a later time that via a separate request )

I wasn't aware previously, but I've checked again, and there are indeed non-blocking alternatives for most use-cases (basically all except the chat ones).

- `onPlayerTryConnect`:
  - ban users -> use `net.banlist_*`  instead
  - VIP slots -> use `net.kick(playerID, message)` instead
- `onPlayerTryChangeSlot`:
  - slot block -> use `net.set_slot(sideID, slotID)` instead

With only the chat use-cases remaining, I'd tend to not support those use-cases and don't merge the PR. We can still log/receive all chat messages, just not modify them.

rurounijones:
> With only the chat use-cases remaining, I'd tend to not support those use-cases and don't merge the PR. We can still log/receive all chat messages, just not modify them.

Yeah, I think that is the best option.

rkusa:
> Is this a POC that is applicable to many use-cases? In this case it is a server-initiated round trip from the server to the network client and back to the server with the time-limit.
> 
> I don't know much about the hook env APIs but this seems like it would be an edge-case? My image is that most of the hook env APIs would be things like streaming data or receiving an incoming request. No time-limited round trips involved.
> 
> (I am not sure if the slot-blocking has the same time-limit block-the-server approach for example. Could we instead receive the slot occupied message then remove the user from it at a later time that via a separate request )

rurounijones:
Restoring the branch so that we can take the code and keep the chat streaming part of things.

PR #37, opened 2021-10-05
implement chat stream

rkusa:
This PR adds chat message streaming (it's a reduced version of https://github.com/DCS-gRPC/rust-server/pull/35 without the filtering).

It can be tested with:

```
.\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{}' 127.0.0.1:50051 dcs.Hook/StreamChat
```

Or via

```
cargo run --example chat_stream
```

The latter example is actually not really useful anymore as the same can be achieved with a simple `grpcurl` call (as it just pints out the received messages right now). The example was previously added to show how to do the filtering. So I am wondering whether we should extend the example a bit or remove it.

Ideas for extending it:
- Add an RPC method for `net.send_chat_to` to allow automatic responses to a player sending a chat message
- Reintroduce message filtering by allowing the caller to deploy arbitrary Lua code as a filter, something like (which filters out all messages starting with a forward slash):

    ```
    .\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{"filter":"if message[0] == '/' then\nreturn ""\nend\nreturn message"}' 127.0.0.1:50051 dcs.Hook/StreamChat
    ```

Wdyt?


rurounijones:
`net.send_chat_to` is definitely something I want to add but I am not sure if it should be done via the hook or mission env. I lean towards using the mission env for everything we can get away with and only use the hook env when no other alternatives present themselves.

Personally I use grpcurl for everything, I am not sure if I would use the examples.

I don't think there is any benefit in deploying lua code to filter messages. It kind of goes against the idea of an RPC system. Just send the messages to the clients and let them filter it client-side in the language of their choice. I don't think there would be much performance saved by server-side filtering.

PR #38, opened 2021-10-12
add option to toggle debug logs (off by default)

rkusa:
Only write `Debug` level logs (eg. logging each event) to `gRPC.log` if `GRPC.debug` is set to `true` (`false` by default).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ GRPC = {
   -- the port to listen on
   port = 50051,
 
-  -- whether debug logging is enabeld or not
+  -- whether debug logging is enabled or not
   debug = false,
 }
 ```
--- a/lua/grpc.lua
+++ b/lua/grpc.lua
@@ -24,9 +24,7 @@ end
 if GRPC.port == nil then
   GRPC.port = 50051
 end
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -87,6 +87,9 @@ GRPC = {
 
   -- the port to listen on
   port = 50051,
+
+  -- whether debug logging is enabeld or not
```
rurounijones, line 91: Nit: typo in `enabled`.

lua/grpc.lua
```diff
@@ -24,6 +24,9 @@ end
 if GRPC.port == nil then
   GRPC.port = 50051
 end
+if GRPC.debug == true then
+  GRPC.debug = true
+end
```
rurounijones, line 29: Isn't this redundant? If a == true then a = true

rkusa: It is indeed not what I intended. I've update it to:

```lua
GRPC.debug = GRPC.debug == true
```

Which still looks redundant, the idea being that I have to make sure that the data type is a boolean to prevent a crash when passing it into the DLL.

PR #57, opened 2021-10-29
add custom.GetMagneticDeclination method

rkusa:
Add `custom.GetMagneticDeclination` method which calculates the magnetic declination for a given lat, lon and alt.

DCS isn't exposing the declination, which is why I've implemented _International Geomagnetic Reference Field (IGRF)_ in [igrf ](https://github.com/rkusa/igrf) for [DATIS](https://github.com/rkusa/DATIS) a while ago. While the calculated declination is not always 100% exactly what DCS uses, it is very close and I'd say close enough (I think the difference is mainly due to my `igrf` using newer source date than DCS).

This PR is mainly a result of the thought that this might be useful for others, too. Wdyt, does this make sense to have as part of DCS-gRPC?

Example:
```
> .\grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs.proto -d '{\"lat\":42.0252750,\"lon\":40.0473825,\"alt\":0.0}' 127.0.0.1:50051 dcs.custom.CustomService/GetMagneticDeclination
{
  "declination": -6.46
}
```

### DIFF
--- a/protos/custom.proto
+++ b/protos/custom.proto
@@ -17,9 +17,11 @@ service CustomService {
   // Disabled by default.
   rpc Eval(EvalRequest) returns (EvalResponse) {}
 
-  /// Calculates the magnetic declination at the given position using the International Geomagnetic
-  /// Reference Field (IGRF) model. The result is not always exactly the same as what DCS seem to
-  /// use, but it is very close (DCS doesn't expose its declination).
+  /**
+   * Calculates the magnetic declination at the given position using the International Geomagnetic
+   * Reference Field (IGRF) model. The result is not always exactly the same as what DCS seem to
+   * use, but it is very close (DCS doesn't expose its declination).
+   */
   rpc GetMagneticDeclination(GetMagneticDeclinationRequest) returns (GetMagneticDeclinationResponse) {}
 }
 
--- a/protos/custom.proto
+++ b/protos/custom.proto
@@ -16,6 +16,11 @@ service CustomService {
   // Evaluate some Lua inside of the mission and return the result as a JSON string.
   // Disabled by default.
   rpc Eval(EvalRequest) returns (EvalResponse) {}
+
+  /// Calculates the magnetic declination at the given position using the International Geomagnetic
+  /// Reference Field (IGRF) model. The result is not always exactly the same as what DCS seem to
+  /// use, but it is very close (DCS doesn't expose its declination).
+  rpc GetMagneticDeclination(GetMagneticDeclinationRequest) returns (GetMagneticDeclinationResponse) {}
 }
 
 message MissionAssignmentRequest {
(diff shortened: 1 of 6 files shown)


src/rpc/mission.rs
```diff
@@ -81,13 +78,25 @@ impl MissionService for MissionRpc {
...
+        .map_err(|err| Status::internal(format!("received invalid day: {}", err)))?;
+    let date = Date::from_calendar_date(year, month, day)
+        .map_err(|err| Status::internal(format!("received invalid date: {}", err)))?;
+    let time = Time::from_hms(0, 0, 0).unwrap() + Duration::seconds(time as i64);
+    let datetime = PrimitiveDateTime::new(date, time).assume_offset(UtcOffset::UTC);
+
+    datetime.format(&Rfc3339).map_err(|err| {
+        Status::internal(format!("failed to format date as ISO 8601 string: {}", err))
+    })
+}
```
rkusa, line 102: `igrf` is based on the `time` crate, which is why I've migrated the code above to the `time` crated too, so that we don't have two dependencies with the same use-case (date/time handling).

rkusa:
Rebased to get fixes from `main` 🔀

protos/custom.proto
```diff
@@ -16,6 +16,11 @@ service CustomService {
   // Evaluate some Lua inside of the mission and return the result as a JSON string.
   // Disabled by default.
   rpc Eval(EvalRequest) returns (EvalResponse) {}
+
+  /// Calculates the magnetic declination at the given position using the International Geomagnetic
+  /// Reference Field (IGRF) model. The result is not always exactly the same as what DCS seem to
+  /// use, but it is very close (DCS doesn't expose its declination).
```
rkusa, line 22: Does `///` work as a doc comment for the doc generator?

src/rpc/custom.rs
```diff
@@ -36,4 +41,36 @@ impl CustomService for MissionRpc {
...
         Ok(Response::new(custom::EvalResponse { json }))
     }
+
+    async fn get_magnetic_declination(
+        &self,
+        request: Request<custom::GetMagneticDeclinationRequest>,
+    ) -> Result<Response<custom::GetMagneticDeclinationResponse>, Status> {
+        let position = request.into_inner();
+        let datetime = self
+            .get_scenario_current_time(Request::new(mission::GetScenarioCurrentTimeRequest {}))
```
rurounijones, line 51: Since these values are not going to change enough to matter over the course of a mission we might consider using the start time instead of the current time.

And since the start time doesn't change is it possible for us to generally call it once at startup and cache the start time when GRPC is started with the mission and then we don't have to call into the MSE at all for it afterwards?

Would save a round-trip to the MSE.

protos/custom.proto
rurounijones: I don't think so.
```c
/*
 *
 */
```
is the multiline format.

rurounijones:
> This PR is mainly a result of the thought that this might be useful for others, too. Wdyt, does this make sense to have as part of DCS-gRPC?

I think this is a great idea. It would be better for clients if they were able to do this themselves of course but not every language might have such a library handy in which case I reckon this is a brilliant fallback / quick option.

src/rpc/custom.rs
rkusa: Good idea, done in a0476b2d575ae68ef85c3583c1d98ac656da69b2 👍

PR #63, opened 2021-11-01
Update coalition enum

rurounijones:
Update the coalition enum to add COALITION_UNSPECIFIED.

We can actually use this new value in some APIs where specifying the
coalition was optional.

I am putting this change up as a PR because it is quite big so want a sanity check.

### DIFF
--- a/protos/dcs/world/world.proto
+++ b/protos/dcs/world/world.proto
@@ -12,7 +12,7 @@ service WorldService {
 }
 
 message GetAirbasesRequest {
-  optional dcs.common.Coalition coalition = 1;
+  dcs.common.Coalition coalition = 1;
 }
 
 message GetAirbasesResponse {
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -19,7 +19,7 @@ GRPC.exporters.unit = function(unit)
     id = tonumber(unit:getID()),
     name = unit:getName(),
     callsign = unit:getCallsign(),
-    coalition = unit:getCoalition(),
+    coalition = unit:getCoalition() + 1, -- Increment for non zero-indexed gRPC enum
     type = unit:getTypeName(),
     position = GRPC.toLatLonPosition(unit:getPoint()),
     playerName = Unit.getPlayerName(unit),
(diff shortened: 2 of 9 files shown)


rurounijones:
As a sidenote: I am not sure why but the test were passing locally when the event Coalition (visibility) on one of the tests was 2 and not 3 but it was passing locally but failing on CI... wierd.

rkusa:
> As a sidenote: I am not sure why but the test were passing locally when the event Coalition (visibility) on one of the tests was 2 and not 3 but it was failing on CI... wierd.

Probably because CI runs `cargo test --workspace`. `cargo test` doesn't test "sub-crates" since I've split the repo into multiple crates.

protos/dcs/world/world.proto
```diff
@@ -14,7 +14,7 @@ service WorldService {
 }
 
 message GetAirbasesRequest {
-  optional dcs.common.Coalition coalition = 1;
+  dcs.common.Coalition coalition = 1; // Use COALITION_UNSPECIFIED to get airbases for all coalitions
```
rkusa, line 17: I wonder whether we should just call it `COALITION_ANY` in the case of the coalition, since it is actively used as a value to _request all_?

rkusa (approved):
👍

protos/dcs/world/world.proto
rurounijones: Lemme see if we can do that and tell the linter to ignore the fact that it doesn't end in `_UNSPECIFIED`

rurounijones: I think `protolint` is probably what we should standardise on in our CI in which case I can make this change using a comment to disable the rule for the line.

Looking at how this is used I think `COALITION_ALL` makes more sense than `COALITION_ANY`. What do you think?

rkusa: `COALITION_ALL` sounds good 👍 
There are two comments about `COALITIONS_UNSPECIFIED` left, LGTM otherwise

rurounijones: Whoops, fixed and remove other references to `unspecified`

PR #69, opened 2021-11-03
add wind vector and move wind calculation to Rust

rkusa:
Points for discussion are highlighted with in-line comments.

### DIFF
--- a/protos/dcs/atmosphere/atmosphere.proto
+++ b/protos/dcs/atmosphere/atmosphere.proto
@@ -28,6 +28,8 @@ message GetWindResponse {
   float heading = 1;
   // The strength of the wind in meters per second
   float strength = 2;
+  // The direction (it is blowing to) and speed (magnitude of the vector) of the wind.
+  dcs.common.Vector velocity = 3;
 }
 
 message GetWindWithTurbulenceRequest {
@@ -40,8 +42,10 @@ message GetWindWithTurbulenceRequest {
 message GetWindWithTurbulenceResponse {
   // The heading the wind is coming from.
   float heading = 1;
-  // The strength of the wind in meters per second
+  // The strength of the wind in meters per second.
   float strength = 2;
+  // The direction (it is blowing to) and speed (magnitude of the vector) of the wind.
+  dcs.common.Vector velocity = 3;
 }
 
 message GetTemperatureAndPressureRequest {
--- a/protos/dcs/atmosphere/atmosphere.proto
+++ b/protos/dcs/atmosphere/atmosphere.proto
@@ -28,8 +28,6 @@ message GetWindResponse {
   float heading = 1;
   // The strength of the wind in meters per second
   float strength = 2;
-  // The direction (it is blowing to) and speed (magnitude of the vector) of the wind.
-  dcs.common.Vector velocity = 3;
 }
 
 message GetWindWithTurbulenceRequest {
@@ -44,8 +42,6 @@ message GetWindWithTurbulenceResponse {
   float heading = 1;
   // The strength of the wind in meters per second.
   float strength = 2;
-  // The direction (it is blowing to) and speed (magnitude of the vector) of the wind.
-  dcs.common.Vector velocity = 3;
 }
 
 message GetTemperatureAndPressureRequest {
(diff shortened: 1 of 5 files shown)


src/rpc/atmosphere.rs
```diff
@@ -31,3 +40,22 @@ impl AtmosphereService for MissionRpc {
...
+        heading -= 180.0;
+    } else {
+        heading += 180.0;
+    }
+
+    // calc 2D strength
+    let strength = (v.z.powi(2) + v.x.powi(2)).sqrt();
+
+    (heading as f32, strength as f32)
+}
```
rkusa, line 61: Wdyt about offloading calculations like that to Rust?

protos/dcs/common/common.proto
```diff
@@ -349,3 +349,14 @@ message MarkPanel {
...
+/**
+ * A vector in a left-handed coordinate system with +z being north, -z south, +x east, -x west,
+ * +y up and -y down (as opposed to DCS' unusual right-handed coordinate system where x is north/
+ * south and z is west/east).
+ */
+ message Vector {
+  double x = 1;
+  double y = 2;
+  double z = 3;
+}
```
rkusa, line 362: I am still not 100% sure whether it is a good idea to change the coordinate system, but we are also changing the wind from a TO direction to a FROM direction - so it is at least not new that we are second guessing DCS.

Pros:
- More common, easier to apply knowledge / code-snippets from general game dev tutorials.

Cons:
- Possibly harder to use for people coming from a DCS Lua background (even though I think that only few people dived into 3D math using DCS vectors).

If the gRPC server wouldn't use the coordinate system as used in this PR, I'd convert vectors on the client-side, which wouldn't be a big deal too.

protos/dcs/atmosphere/atmosphere.proto
```diff
@@ -40,8 +42,10 @@ message GetWindWithTurbulenceRequest {
 message GetWindWithTurbulenceResponse {
   // The heading the wind is coming from.
   float heading = 1;
-  // The strength of the wind in meters per second
+  // The strength of the wind in meters per second.
   float strength = 2;
+  // The direction (it is blowing to) and speed (magnitude of the vector) of the wind.
+  dcs.common.Vector velocity = 3;
```
rkusa, line 48: I actually don't need the vector in the response for my own use-case anymore. So I'd be fine with not including it in the response, but I was wondering whether there might still be use-cases for the velocity in 3D space rather then the strength in 2D. We could also remove it again, until the need comes up again. Wdyt?

src/rpc/atmosphere.rs
rurounijones: Pros:
* It is more efficient I imagine
* Moves stuff out of the MSE
Cons:
* You now need to know Rust instead of Lua

I think it is a great idea for code we are sure will not need much modification (Especially if that modificaiton is likely to be done by someone who is more familiar with Lua). I think this is a good example.

protos/dcs/common/common.proto
rurounijones: I am fine changing the coordinate system at the gRPC boundary. This is sort of an extension of our decision a wile ago not to be slaves to the ED APIs.

However I would extensively document this to make it clear that we are doing it and what the values would be in the lua scripting environment. (e.g. y = z or whatever).

protos/dcs/atmosphere/atmosphere.proto
rurounijones: I cannot think of any use-case right now where a client would need this information and since it is the opposite of `heading` (From and To) I would leave it out for the moment.

If we leave it in then we should add more documentation around these two values and why they are opposite.

protos/dcs/common/common.proto
rkusa: Extended the comment slightly, let me know if you'd like to see more explanations

protos/dcs/atmosphere/atmosphere.proto
rkusa: Removed it 👍

rurounijones (approved):

protos/dcs/atmosphere/atmosphere.proto
rurounijones: I may be missing something but I don't see them removed?

rkusa: `git push` helps 🤦‍♂️

rurounijones (approved):

PR #73, opened 2021-11-10
add category to common.v0.Unit

rkusa:
I had it a second time today, that I had to fetch the group of a unit just to get it's category, which is why I'd propose to always add the category to the units.

Any objections?

### DIFF
--- a/lua/exporters/object.lua
+++ b/lua/exporters/object.lua
@@ -45,7 +45,8 @@ GRPC.exporters.unit = function(unit)
     groupName = Unit.getGroup(unit):getName(),
     numberInGroup = unit:getNumber(),
     heading = heading,
-    speed = speed
+    speed = speed,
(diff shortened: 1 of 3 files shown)


lua/exporters/object.lua
```diff
@@ -45,7 +45,8 @@ GRPC.exporters.unit = function(unit)
     groupName = Unit.getGroup(unit):getName(),
     numberInGroup = unit:getNumber(),
     heading = heading,
-    speed = speed
+    speed = speed,
+    category = unit:getGroup():getCategory(),
```
rkusa, line 49: There is also a `unit:getCategory()` but that seems to be deprecated as it always returns `0`

rkusa:
Fixed the unit tests

rurounijones (approved):

PR #75, opened 2021-11-11
add `autostart` option

rkusa:
Add `autostart` option which automatically loads and starts the gRPC server inside of each mission. When set, it is not necessary to manually add the server to each mission.

See [`README`](https://github.com/DCS-gRPC/rust-server/blob/autostart/README.md) for usage, please let me know if the README is not clear enough.


rurounijones (approved):
Nice!

PR #77, opened 2021-11-13
Airplane Group Spawning

Wizxrd:
Initial implementation of spawning an airplane group from an airbase, zone, or lat lon coordinate with any number of points. When spawning from an airbase you have the option to spawn hot, cold, runway and in air.
Individual unit payloads are yet to be implemented as are taskings for each point in the route. These can be complex tasks to get done properly, so need to think about how it should be done easily for the end user.

added createPlaneGroupTemplate
returns a table required to spawn an airplane group

added createPlaneGroupUnitsTemplate
return a table of units needed to spawn an airplane
prepares units to spawn from an airbase, zone, or lat lon
if spawned from an airbae options are hot, cold, air, or runway

added buildAirplanePoints
return a table of built points for airplane groups
points are based around vec3's that can be from zones, airbases and lat lon

**notes:**
- when spawning an airplane from an airbase that is a ship it is required to use the unit name as the place
- the place within units is the inital spawning place of the units
- changed the addGroup method to include being able to pick a differnt tempalte based on params.template.type
- types could be "Ground" "Helicopter, "Airplane", "Sea" if implemented that way

### DIFF
--- a/lua/methods/coalitions.lua
+++ b/lua/methods/coalitions.lua
@@ -256,7 +256,7 @@ GRPC.methods.addGroup = function(params)
   end
   local template
   if params.template.type == "Airplane" then
-    template = createPlaneGroupTemplate(params.template.groupTemplate)
+    template = createPlaneGroupTemplate(params.template.airplaneTemplate)
   elseif params.template.type == "Ground" then
     template = createGroundGroupTemplate(params.template.groundTemplate)
   end
--- a/lua/methods/coalitions.lua
+++ b/lua/methods/coalitions.lua
@@ -21,12 +21,157 @@ local skill = {
   Player = 5
 }
 
+-- types for aircraft taking off
+local takeoffTypes = {
+  [1] = "TakeOff",
+  [2] = "TakeOffParking",
+  [3] = "TakeOffParkingHot",
+  [4] = "Turning Point",
+  runway = "TakeOff",
+  cold = "TakeOffParking",
+  hot = "TakeOffParkingHot",
+  air = "Turning Point"
+}
+
+-- actiions for aircraft taking off
+local takeoffActions = {
+  [1] = "From Runway",
+  [2] = "From Parking Area",
+  [3] = "From Parking Area Hot",
+  [4] = "Turning Point",
+  runway = "From Runway",
+  cold = "From Parking Area",
+  hot = "From Parking Area Hot",
+  air = "Turning Point",
+}
+
 --local altitudeType = {
 --  [1] = "BARO",
 --  [2] = "RADIO",
 --  BARO = 1,
 --  RADIO = 2
 --}
+
+local buildAirplanePoints = function(points)
+  local builtPoints = {}
+  for _, pointData in pairs(points) do
+    local pointVec3
+    if type(pointData.place) == "string" then
+      if Airbase.getByName(pointData.place) then
+        pointVec3 = Airbase.getByName(pointData.place):getPoint()
+        env.info("ab point type")
+      elseif trigger.misc.getZone(pointData.place) then
+        pointVec3 = trigger.misc.getZone(pointData.place).point
+        env.info("zone point type")
+      end
+    elseif type(pointData.place) == "table" then
+      pointVec3 = coord.LLtoLO(pointData.place.lat, pointData.place.lon)
+    end
+    builtPoints[#builtPoints+1] = {
+      alt = pointData.alt,
+      x = pointVec3.x,
+      y = pointVec3.z,
+      type = pointData.type or takeoffTypes.air,
+      eta = 0,
+      eta_locked = true,
+      alt_type = pointData.alt_type or "BARO",
+      formation_template = "",
+      speed = pointData.speed,
+      action = pointData.action or takeoffActions.air,
+      task = {
+        id = "ComboTask",
+        params = {
+          tasks = {}
+        }
+      }
+    }
+  end
+  -- here we ammend the first point to allow for spawns from airbases if it isn't an airspawn
+  if Airbase.getByName(points[1].place) then
+    if points[1].type ~= "Turning Point" and points[1].action ~= "Turning Point" then
+      local ab = Airbase.getByName(points[1].place)
+      local abId = Airbase.getID(ab)
+      local abCat = Airbase.getDesc(ab).category
+      if abCat == 0 then -- Airbase.Category.AIRDROME
+        builtPoints[1].airdromeId = abId
+      elseif abCat == 2 then -- Airbase.Category.HELIPAD
+        builtPoints[1].linkUnit = abId
+        builtPoints[1].helipadId = abId -- why its named helipad i dont know
+      end
+    end
+  end
+  return builtPoints
+end
+
+local createPlaneGroupUnitsTemplate = function(unitListTemplate)
+    local units = {}
+    for i, unitTemplate in pairs(unitListTemplate) do
+      local pointVec3
+      if type(unitListTemplate.place) == "string" then
+        if Airbase.getByName(unitListTemplate.place) then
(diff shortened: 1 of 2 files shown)


lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+          env.info("ab point type")
+        elseif trigger.misc.getZone(unitListTemplate.place) then
+          pointVec3 = trigger.misc.getZone(unitListTemplate.place).point
+          env.info("zone point type")
+        end
+      elseif type(unitListTemplate.place) == "table" then
+        pointVec3 = coord.LLtoLO(unitListTemplate.place.lat, unitListTemplate.place.lon)
+      end
+      local fuel = Unit.getDescByName(unitListTemplate.type).fuelMassMax -- needed incase no payload table is applied
+      if type(unitTemplate) == "table" then
```
rurounijones, line 122: This will always be true so we don't need the `if` here.

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+          ["pylons"] = {},
+          ["fuel"] = fuel,
+          ["flare"] = 0,
+          ["chaff"] = 0,
+          ["gun"] = 0,
+          },
+          parking = unitTemplate.parking or nil,
+          parking_id = unitTemplate.parking_id or nil,
+          callsign = unitTemplate.callsign or nil,
+          skill = unitTemplate.skill or skill.Random,
```
rurounijones, line 141: Change to match the groundUnit `skill = skill[unitTemplate.skill]`

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+-- types for aircraft taking off
+local takeoffTypes = {
+  [1] = "TakeOff",
+  [2] = "TakeOffParking",
+  [3] = "TakeOffParkingHot",
+  [4] = "Turning Point",
+  runway = "TakeOff",
+  cold = "TakeOffParking",
+  hot = "TakeOffParkingHot",
+  air = "Turning Point"
```
rurounijones, line 33: We don't need these for the moment so lets remove them until we have a definite need.

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+-- actiions for aircraft taking off
+local takeoffActions = {
+  [1] = "From Runway",
+  [2] = "From Parking Area",
+  [3] = "From Parking Area Hot",
+  [4] = "Turning Point",
+  runway = "From Runway",
+  cold = "From Parking Area",
+  hot = "From Parking Area Hot",
+  air = "Turning Point",
```
rurounijones, line 45: We don't need these for the moment so lets remove them until we have a definite need.

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
   Player = 5
 }
 
+-- types for aircraft taking off
+local takeoffTypes = {
```
rurounijones, line 25: Ground units also use the "Turning Point" type. I think this would better be named as "waypointType" (also, singular)

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+  [3] = "TakeOffParkingHot",
+  [4] = "Turning Point",
+  runway = "TakeOff",
+  cold = "TakeOffParking",
+  hot = "TakeOffParkingHot",
+  air = "Turning Point"
+}
+
+-- actiions for aircraft taking off
+local takeoffActions = {
```
rurounijones, line 37: Same as above so this would be `waypointAction`

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
 --  [2] = "RADIO",
 --  BARO = 1,
 --  RADIO = 2
 --}
 
+local buildAirplanePoints = function(points)
+  local builtPoints = {}
+  for _, pointData in pairs(points) do
+    local pointVec3
+    if type(pointData.place) == "string" then
```
rurounijones, line 59: I'll think about if we can do this in a different way by having gRPC types. No need to change this for the moment, just an FYI.

If may be that we do not allow clients to specify anything other than a Lat/Lon/Alt point since the client can get the information on airfield locations and zones themselves and send us that data instead of making us find it server-side.

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+      if Airbase.getByName(pointData.place) then
+        pointVec3 = Airbase.getByName(pointData.place):getPoint()
+        env.info("ab point type")
```
rurounijones, line 62: Can you remove these log statements. They do not provide much actionable information and we have a separate logger for gRPC anyway. We should not be logging to the env.

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+      elseif trigger.misc.getZone(pointData.place) then
+        pointVec3 = trigger.misc.getZone(pointData.place).point
+        env.info("zone point type")
```
rurounijones, line 65: Remove

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+      pointVec3 = coord.LLtoLO(pointData.place.lat, pointData.place.lon)
+    end
+    builtPoints[#builtPoints+1] = {
+      alt = pointData.alt,
+      x = pointVec3.x,
+      y = pointVec3.z,
+      type = pointData.type or takeoffTypes.air,
+      eta = 0,
+      eta_locked = true,
+      alt_type = pointData.alt_type or "BARO",
```
rurounijones, line 77: Can you switch this to using https://github.com/DCS-gRPC/rust-server/blob/main/lua/methods/coalitions.lua#L24 (and uncomment it, I only commeneted it because it was not being used yet and making the lint checker complain)

lua/methods/coalitions.lua
```diff
@@ -21,13 +21,158 @@ local skill = {
...
+      task = {
+        id = "ComboTask",
+        params = {
+          tasks = {}
+        }
+      }
+    }
+  end
+  -- here we ammend the first point to allow for spawns from airbases if it isn't an airspawn
+  if Airbase.getByName(points[1].place) then
```
rurounijones, line 90: Again as an FYI this bit may change, but nothing to do yet.

rurounijones (changes requested):
Been talking to Wizxrd in Discord about this looking at this.

Added some comments that could be fixed now while I ponder what we can do in the gRPC layer to simplify things.

lua/methods/coalitions.lua
rurounijones: On reflection I think we should only allow Lat/Lon/alt using the `Position` Protobuf message. Clients should be able to get all the information they need to be able to popualte this information themselves and it simplifies the lua a lot.

justin-lovell:
Howdy @rurounijones  and @Wizxrd ! I am interested in this feature being added to the core project; and I can't find Discord link anywhere so I can't follow up on that channel -- is there something I could do to help with this first push, and follow up with additional work? E.g. Ships.

rurounijones:
We are using the OverlordBot Discord for this (Don't want to setup another project specific Discord). https://discord.gg/9RqyTJt

I am busy with OverlordBot stuff at the moment, but there was a bit of discussion in the #developers channel of the gRPC category of that Discord. So you can have a look there.

Basically it was, as far as my memory goes, what changes we can make to this to abstract some things at the gRPC level using existing constructs (e.g. using a `Location` construct instead of sending lat/lon at the gRPC level)

PR #85, opened 2021-12-22
Add logging around config file writing

rurounijones:
Add logging regarding the writing of the `Data\dcs-grpc.lua`
file and change the check to make sure that `f` is not nill rather
than there is a definite error as this appears to be the safer, more
idiomatic, way of doing it.

This should help a bit with debugging issues some users appear to
be having.

### DIFF
--- a/lua/grpc-mission.lua
+++ b/lua/grpc-mission.lua
@@ -15,18 +15,19 @@ end
 -- Let DCS know where to find the DLLs
 package.cpath = package.cpath .. GRPC.dllPath .. [[?.dll;]]
 
+env.info("[GRPC] Writing " .. lfs.writedir() .. [[Data\dcs-grpc.lua]] )
 -- Make settings available to gRPC hook
-local file, err = io.open(lfs.writedir() .. [[Data\dcs-grpc.lua]], "w")
-if err then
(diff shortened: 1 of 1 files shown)


lua/grpc-mission.lua
```diff
@@ -15,18 +15,19 @@ end
...
 
+env.info("[GRPC] Writing " .. lfs.writedir() .. [[Data\dcs-grpc.lua]] )
 -- Make settings available to gRPC hook
-local file, err = io.open(lfs.writedir() .. [[Data\dcs-grpc.lua]], "w")
-if err then
-	env.error("[GRPC] Error writing config")
-else
-	file:write(
+local f, err = io.open(lfs.writedir() .. [[Data\dcs-grpc.lua]], "w")
+if f then
```
rkusa, line 21:
```suggestion
local file, err = io.open(lfs.writedir() .. [[Data\dcs-grpc.lua]], "w")
if file then
```

or rename `file` below to `f` 🙂

rkusa (changes requested):

lua/grpc-mission.lua
rurounijones: Thanks, that is what I get for early morning commits :cry:

PR #87, opened 2021-12-23
add support for config file

rkusa:
Add support to configure the gRPC server via a config file. This should help users that run multiple servers in parallel and thus need different settings per instance.

See [README](https://github.com/DCS-gRPC/rust-server/blob/config-file/README.md#settings) for instructions of how the config file works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,21 +83,36 @@ GRPC.throughputLimit = 600
 GRPC.autostart = false
 ```
 
-Settings can either be set above `GRPC.load()` inside of a mission, or globally inside of the `MissionScripting.lua`, e.g.:
+Settings can either be set:
+- above `GRPC.load()` inside of a mission,
+  ```lua
+  GRPC.debug = true
+  GRPC.autostart = true
+  GRPC.load()
+  ```
 
-```diff
-  --Initialization script for the Mission lua Environment (SSE)
+- inside a `Saved Games\DCS\Config\dcs-grpc.lua` file (useful if run multiple servers in parallel), or
 
-  dofile('Scripts/ScriptingSystem.lua')
-+
-+ GRPC = {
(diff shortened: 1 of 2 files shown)


lua/grpc-mission.lua
```diff
@@ -1,7 +1,18 @@
--- Set default settings.
 if not _G.GRPC then
   _G.GRPC = {}
 end
+
+-- load settings from `Saved Games/DCS/Config/dcs-grpc.lua`
+do
```
rurounijones, line 6: Can we add some logging like.

* "Checking for DCS/Config/dcs-grpc.lua file"
* "DCS/Config/dcs-grpc.lua file not found"

To help with debugging?

README.md
```diff
@@ -83,21 +83,36 @@ GRPC.throughputLimit = 600
...
+- above `GRPC.load()` inside of a mission,
+  ```lua
+  GRPC.debug = true
+  GRPC.autostart = true
+  GRPC.load()
+  ```
+
+- inside a `Saved Games\DCS\Config\dcs-grpc.lua` file (useful if run multiple servers in parallel), or
+
+  `Saved Games\DCS\Config\dcs-grpc.lua`:
```
rkusa, line 96: Wdyt about the name?
Other naming options would be: `grpc.lua`, `DCS-gRPC.lua`

lua/grpc-mission.lua
rkusa: Done in 8cbf23ab6df327329a3a057aa87d5b61a5b44216 👍

README.md
rurounijones: Name is fine I think. No strong opinions on alternatives.

rurounijones (approved):

PR #88, opened 2021-12-23
add connect and disconnect events

### LINKED ISSUE #84: Add client connect / disconnect to the event stream
Adapt https://github.com/mmiller84/rust-server/commit/8fcbc84f0cbb4da20ba4bf07026441b8d59d687f but instead of having a separate API, add them as events into the existing event stream we use for all the DCS events.

rkusa:
Add `connect` and `disconnect` events.

Example:

```
2021-12-23 10:34:36.594 DEBUG   dcs_grpc: Received event: StreamEventsResponse {
    time: 8.341,
    event: Some(
        Connect(
            ConnectEvent {
                addr: "x.x.x.x:51183",
                name: "Binary",
                ucid: "...",
                id: 2,
            },
        ),
    ),
}
2021-12-23 10:35:18.196 DEBUG   dcs_grpc: Received event: StreamEventsResponse {
    time: 49.959,
    event: Some(
        Disconnect(
            DisconnectEvent {
                id: 2,
                reason: ThatsOkay,
            },
        ),
    ),
}
```

Fixes #84


protos/dcs/mission/v0/mission.proto
```diff
@@ -343,6 +386,8 @@ message StreamEventsResponse {
     // @exclude 38 reserved for S_EVENT_TRIGGER_ZONE
     LandingQualityMarkEvent landing_quality_mark = 39;
     // @exclude 40 reserved for S_EVENT_BDA
+    ConnectEvent connect = 41;
```
rkusa, line 389: Wdyt, should we continue with the next number here, or leave some field numbers free for actual future DCS events?

protos/dcs/mission/v0/mission.proto
```diff
@@ -303,6 +303,49 @@ message StreamEventsResponse {
...
+   * The reason a player disconnected for.
+   */
+  enum DisconnectReason {
+    DISCONNECT_REASON_THATS_OKAY = 0;
+    DISCONNECT_REASON_INVALID_ADDRESS = 1;
+    DISCONNECT_REASON_CONNECT_FAILED = 2;
+    DISCONNECT_REASON_WRONG_VERSION = 3;
+    DISCONNECT_REASON_PROTOCOL_ERROR = 4;
+    DISCONNECT_REASON_TIMEOUT = 5;
+    DISCONNECT_REASON_INVALID_PASSWORD = 100;
```
rkusa, line 330: Do you still have a proto linter installed locally? If so, is the linter unhappy with the non-continuous numbering here (going from `5` to `100`)?

protos/dcs/mission/v0/mission.proto
rurounijones: I think it is safer for us to chose a large random number. like 29583 or something that is extremely unlikely to be used by DCS itself.

protos/dcs/mission/v0/mission.proto
rurounijones: Non-contiguous numbers is fine but it will complain about the use of 0. It expects

`DISCONNECT_REASON_INVALID = 0` is what it will expect.

I recommend shifting everything up by 1 like we did elsewhere

protos/dcs/mission/v0/mission.proto
rkusa: Do you know if this increases the size of the protobuf message?

protos/dcs/mission/v0/mission.proto
rkusa: Hm, I kinda liked the default to be `0`, which is the user leaving the server without an "error". How'd you feel about just renaming it to `DISCONNECT_REASON_UNSPECIFIED` and keeping the meaning of "manual disconnect" (aka no error)?

rurounijones: My worry is that using 0 will cause a client who happens to use Language X extra pain in the future for no real technical reason. `Unspecified` is not really the same as `No error` so I am not entirely sure we can get away with calling them the same thing.

To avoid this potential issue requires a `reason + 1` to line 88 of the hook lua and incrementing the values here so I don't think we can claim a solid technical reason for not doing it.

It will also improve consistency since we have changed all the other ENUMs in the codebase to use 0 as unspecified.

protos/dcs/mission/v0/mission.proto
rurounijones: It doesn't.

protos/dcs/mission/v0/mission.proto
justin-lovell: What if we call it `DISCONNECT_REASON_UNKNOWN` or `DISCONNECT_REASON_UNSPECIFIED`? The latter is probably more correct with the DCS API because users can just - you know - leave the server ( insert drum roll -- I'll show myself out :D )

rurounijones: `DISCONNECT_REASON_UNSPECIFIED` and `DISCONNECT_REASON_THATS_OKAY` (i.e. no error) are not synonymous to me, which is what my previous comment was trying to get at.

protos/dcs/mission/v0/mission.proto
rkusa: Done in a2993aeb2e3cd6a7f57f5d5f015b2f468e940509. I didn't go thaaatt high, my mind felt better if I could still spot that the number is a multiple of 1024

protos/dcs/mission/v0/mission.proto
rkusa: Done in c869bea9d67b289ba3f85910705b2d5b9b93e1f6 👍

rurounijones (approved):

rkusa:

PR #89, opened 2021-12-29
Chat messages are synchronized to event stream

justin-lovell:
I am planning to build features on top of synchronized chronological order of events. Although we are getting the messages sent as a stream, there is disadvantage of synchronizing the two streams due to no timestamp being emitted; or alternatively use a Rx function with a lag/latency con.

Taking inspiration of this PR, we should emit the chat messages to the single stream: https://github.com/DCS-gRPC/rust-server/pull/88

I've preempted some feedback by reviewing the above PR. Let me know if you have any strong opinions.

Note: the existing chat stream will still be populated.

Updated with sample output.
```

2022-01-06 19:15:51.674 DEBUG   dcs_grpc: Received event: StreamEventsResponse {
    time: 65.261,
    event: Some(
        PlayerSendChat(
            PlayerSendChatEvent {
                player_id: 2,
                message: "this is to the world",
                target: All,
            },
        ),
    ),
}

2022-01-06 19:15:57.809 DEBUG   dcs_grpc: Received event: StreamEventsResponse {
    time: 71.397,
    event: Some(
        PlayerSendChat(
            PlayerSendChatEvent {
                player_id: 2,
                message: "and to my allies",
                target: All,
            },
        ),
    ),
}
```


lua/grpc-hook.lua
```diff
@@ -62,6 +62,20 @@ local function init()
 
   function handler.onPlayerTrySendChat(playerID, msg, all)
     _G.GRPC.onChatMessage(playerID, msg, all)
+
+    local target = 0
+    if all then
+      target = 1
+    end
```
justin-lovell, line 69: Note: DCS events are always emitting `all=true` -- so this is a simulation related bug, unfortunately.

protos/dcs/mission/v0/mission.proto
```diff
@@ -303,6 +303,20 @@ message StreamEventsResponse {
     dcs.common.v0.Airbase place = 3;
   }
 
+  // Occurs when a chat message is sent on the server
+  message PlayerSentChatEvent {
+    // The player's id in the current server session.
+    uint32 playerSessionId = 1;
```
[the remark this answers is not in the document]
rurounijones: I think including the chat messages into the main event stream is fine and makes perfect sense.

Regarding consistency.
1. Yes our gRPC APIs should be consistent in their field naming whereever possible.
2. We decided when we started the project that we would not be slaves to EDs naming scheme and would use our own API names & our own field names if we thought the ED naming scheme was not very good.  A common case was redundancy in the field names redundancy in the field names.

So consistency in _our_ public API: Yes but sticking to ED naming: No. I probably need to write that down somewhere

As a test case. Here we have `playerSessionId` which is part of the `PlayerSentChatEvent`. We also have an `id` field in the `PlayerConnectEvent` in https://github.com/DCS-gRPC/rust-server/pull/88

I am assuming they mean the same thing here and therefore should be the same. `id` could be argued to be ambiguous (player id? message id? connection id?) however I think `playerSessionId` is also redundant since this is an event _about_ a player. Therefore `sessionId` could be adequate.

However if there are other events that are not specific to being about "A player" then `sessionId` could be ambiguous as well.

I will let others weigh in on which name we should use here.

lua/grpc-hook.lua
rurounijones: I would say "report it to ED" but that would probably be a waste of time.

rurounijones (changes requested):
Thanks for the contribution!

Could you update the Commit message to include a body with more information. e.g. 

Something along the lines of what you wrote in the Github summary should go into the commit message.

```
50 character summary

Commit message body. Each line limited to 72'ish characters. URLs and other strings that cannot be broken up being an exception. Note the blank line between summary and body.
```

justin-lovell (commented):
@rurounijones, I've updated the commit message and preempted the revision of the `playerSessionId` to `playerId` based on evidence submitted on that local thread.

rurounijones (approved):

protos/dcs/mission/v0/mission.proto
```diff
@@ -303,6 +303,20 @@ message StreamEventsResponse {
     dcs.common.v0.Airbase place = 3;
   }
 
+  // Occurs when a chat message is sent on the server
+  message PlayerSentChatEvent {
```
rkusa, line 307: Would `PlayerSendChatEvent` (`Send` instead of `Sent`) maybe a bit more consistent naming compared to the other event names? Wdyt?

protos/dcs/mission/v0/mission.proto
```diff
@@ -303,6 +303,20 @@ message StreamEventsResponse {
...
+    // The player's id in the current server session.
+    uint32 playerId = 1;
+    // what was typed
+    string message = 2;
+    // to whom?
+    MessageTarget target = 3;
+    enum MessageTarget {
+      COALITION = 0;
```
rkusa, line 315: @rurounijones you know if the protobuf linter would also complain here that the `0` entry its not named `UNSPECIFIED`?

rkusa (commented):
Adding this as a custom event is way better than the extra stream we have now 👍. I'd actually be fine with removing the chat stream in favour of this event.

protos/dcs/mission/v0/mission.proto
justin-lovell: My habit from other environments leaked here -- an "event" should as past tense; I didn't triple check the consistency on other events. 

If we are trying to convince other developers to move over to this dependency compared to the in-game stuff, yeah, keeping it close to the original DCS API scheme (active over past) would make sense.

I'll update that name -- and during an API review we want to change the convention, we uplift everything at the same time.

justin-lovell (commented):
> I'd actually be fine with removing the chat stream in favour of this event.

Yeah, I agree. Have you formulated a deprecation strategy?
(I know we are v0 but an instantaneous change is not always welcomed).

rkusa:
> Have you formulated a deprecation strategy? (I know we are v0 but an instantaneous change is not always welcomed).

I don't think we have a deprecation strategy yet. While I don't have a strong opinion, I'd slightly tend towards just removing it as long as we are `v0` 🤷‍♂️

protos/dcs/mission/v0/mission.proto
rkusa: I think https://github.com/DCS-gRPC/rust-server/pull/89/commits/de2be5b7ab35969973fbc4d10163478aa7848c6b is missing the following change:

```suggestion
  message PlayerSendChatEvent {
```

protos/dcs/mission/v0/mission.proto
```diff
@@ -392,6 +406,7 @@ message StreamEventsResponse {
     // 8192 to give DCS plenty of space for new built-in events.
     ConnectEvent connect = 8192;
     DisconnectEvent disconnect = 8193;
+    PlayerSentChatEvent player_sent_chat = 8194;
```
rkusa, line 409: I think https://github.com/DCS-gRPC/rust-server/pull/89/commits/de2be5b7ab35969973fbc4d10163478aa7848c6b is missing the following change:

```suggestion
    PlayerSendChatEvent player_send_chat = 8194;
```

protos/dcs/mission/v0/mission.proto
rurounijones: Good catch!. Yes same issue and same resolution. The enum field names also need to contain the enum name as follows:

```proto
enum MessageTarget {
  MESSAGE_TARGET_UNSPECIFIED = 0
  MESSAGE_TARGET_COALITION = 1
  MESSAGE_TARGET_ALL = 2
}
```

Then + 1 in the lua with a comment.

protos/dcs/mission/v0/mission.proto
justin-lovell: Thanks. Addressed in https://github.com/DCS-gRPC/rust-server/pull/89/commits/bc90a094a394f68db7b6b327dffa23134b86d96d

protos/dcs/mission/v0/mission.proto
justin-lovell: Addressed in https://github.com/DCS-gRPC/rust-server/pull/89/commits/0e60ac92fb4bc2fd8f3e558c3e6efbef4532e4b9

justin-lovell:
> I'd slightly tend towards just removing it as long as we are v0 

@rkusa 
Fenced In. Pickle. Shack. Damage report: https://github.com/DCS-gRPC/rust-server/pull/89/commits/72aa9dd90102486a960e3ce8f89ed045ff8a0c83

protos/dcs/mission/v0/mission.proto
```diff
@@ -303,6 +303,21 @@ message StreamEventsResponse {
...
+    uint32 playerId = 1;
+    // what was typed
+    string message = 2;
+    // to whom?
+    MessageTarget target = 3;
+    enum MessageTarget {
+      MESSAGE_TARGET_UNSPECIFIED = 0;
+      MESSAGE_TARGET_COALITION = 1;
+      MESSAGE_TARGET_ALL = 2;
+    }
```
rkusa, line 318: Nit: Just noticed, while seeing the removed code form the previous chat stream that the enum could also be covered by

```
  // Whether the message was send to all or only to allies.
   bool all = 4;
```

But I don't have a strong opinion here, I am fine with either one.

rkusa (approved):
👍

protos/dcs/mission/v0/mission.proto
rurounijones: It also occurs to me.

With the current setup the client will now know what coalition the message is sent to based on this message. They would need to go and get the player information and infer from there.

Should we instead enumerate the coalitions instead and have the server determine what the value should be based on the player? 

If we do we could reuse the coalition enum from `common`. https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/common/v0/common.proto#L43

justin-lovell: I like that idea @rurounijones. I would implement it in a heartbeat except that the `all` flag from DCS API is bugged. See my note here: https://github.com/DCS-gRPC/rust-server/pull/89#discussion_r776235960

Perhaps at this moment, knowing we have a bug which would not activate the cool API, should we just skip this field? I'll leave a comment with further extension when the DCS environment behaves correctly.

rurounijones: That is a good point. I think skipping the field entirely is probably the better option as it means no backwards compatibility issues if ED every fix it and we choose to add it since there would be no existing field that needs modification.

rkusa: +1 for skipping it for now

PR #90, opened 2021-12-29
more development hints for hook workflow

justin-lovell:
When developing and expanding upon the `net` based events such as [onChatMessage](https://github.com/DCS-gRPC/rust-server/pull/89) and [onPlayerTryConnect](https://github.com/DCS-gRPC/rust-server/pull/88), developers will need to have a slight better way to expand the hooks available.


rurounijones:
I am not particularly up-to-speed on hook stuff so I will leave this one to @rkusa

rkusa:
I am afraid that I don't understand what you mean with

> developers will need to have a slight better way to expand the hooks available

@justin-lovell do you have an example for what you'd say developers will need?

lua/grpc-hook.lua
```diff
@@ -1,4 +1,13 @@
...
+if not _G.GRPC then
+  _G.GRPC = {}
+end
+
+if not GRPC.dllPath then
+  GRPC.dllPath = lfs.writedir() .. [[Mods\tech\DCS-gRPC\]]
+end
+
+-- Let DCS know where to find the DLLs
+package.cpath = package.cpath .. GRPC.dllPath .. [[?.dll;]]
```
rkusa, line 10: I don't think that this is necessary. The current hook already picks up the dll und lua paths you set up in your `MissionScripting.lua`. It works as follows:

- You set the `dllPath` and `luaPath` in the `MissionScripting.lua`
- They are written into `Data\dcs-grpc.lua` (temporary file to exchange settings between the mission and hook env)
- `grpc-hook.lua` reads `Data\dcs-grpc.lua` https://github.com/DCS-gRPC/rust-server/blob/main/lua/grpc-hook.lua#L8
-  ... and uses `dllPath` and `luaPath` from `Data\dcs-grpc.lua` to set `package.cpath`

However, after looking at it again, I think there is indeed a bug. Simply removing https://github.com/DCS-gRPC/rust-server/blob/main/lua/grpc-hook.lua#L1 might fix the whole setup already.

rkusa (changes requested):
Thanks for the PR! I think I understand what you are trying to achieve. This is something that is supposed to already work. I think I also spotted why it isn't. I've added an inline comment to explain it. Can you maybe give the suggested fix a try to see if this solves the dev setup?

lua/grpc-hook.lua
justin-lovell: So, I've understood there are two entry-points to make things "work":

1. `MissionScripting.lua` which would allow mission based events (deaths, get groups, births, ect.) to be emitted. The temporary redirect to a local development build is already described in the docs.
2. `Hooks/dcs-grpc.lua` which would allow the interception of the `net.*` based API's (chat messages, players changing slots, ect.). This is missing.

One of my first attempts to change the DCS instance to load up my WIP on the `net.*` level was to change that line to `GRPC.dllPath`. This was undefined.

So these changes are necessary, and the duplication of the file paths was disappointing. But it was necessary to redirect to my local build for rapid testing.

However, on a fresh mind, I think I have a better alternative by scraping the `MissionScripting.lua` modifications completely. This should greatly simplify installation instructions!

justin-lovell: > However, on a fresh mind, I think I have a better alternative by scraping the MissionScripting.lua modifications completely. This should greatly simplify installation instructions!

And I just remembered why `<saved games>/DCS/Scripts/hooks` don't play nice with mission scoped variables. At least I got to a deeper understanding on a few other things.

> However, after looking at it again, I think there is indeed a bug. Simply removing https://github.com/DCS-gRPC/rust-server/blob/main/lua/grpc-hook.lua#L1 might fix the whole setup already.

Yeah, that's right -- I discovered that on my little experiment earlier on. And I've updated that, although that was not the primary issue I was experiencing.

---

But there still needs to be an initial "hook" which will bootstrap the `Development/lua/grpc-hook.lua`.

I'm 100% certain that this is a missing hook advice (as per the `README`) is still necessary; I've since then simplified this part. An example API which will never fire whilst in development is:

https://github.com/DCS-gRPC/rust-server/blob/315090af97b6821e969859864485e7d93aa755a8/lua/grpc-hook.lua#L43-L48

The changes which I performed in my PR was not being reflected -- hence the requirement to "hot load" the in development `grpc-hook.lua`: https://github.com/DCS-gRPC/rust-server/pull/89/files#diff-aad7a754803d5073771ebf46890314bfe01603b9ef0377f1b63e397d3af3f92aR65-R78

lua/grpc-hook.lua
```diff
@@ -18,7 +18,7 @@ local function load()
   end
 
   -- Let DCS know where to find the DLLs
-  if not string.find(package.cpath, "DCS-gRPC") then
+  if not string.find(package.cpath, GRPC.dllPath) then
```
justin-lovell, line 21: BTW, this would prevent another bug when working with development paths.

lua/grpc-hook.lua
rkusa: Looks good, I have a suggestion I am going to make, but I'll have to get some other (related) changes out of the way first. I'll get back to your PR once I have my suggestion ready.

rkusa: I'd suggest this instead: https://github.com/DCS-gRPC/rust-server/pull/95

justin-lovell: The main resolution of this PR is not addressed in #95 -- the `Saved Games/DCS/Scripts/Hooks/` still needs to call the development branch.

See these specific lines: https://github.com/DCS-gRPC/rust-server/blob/f6709806bf77292eb7ed06d3feec951f789324ba/README.md?plain=1#L176-L190

rkusa: It does in #95. Once you set the `luaPath` to point to your dev directory in the config, it will load `dcs-hook.lua` from there https://github.com/DCS-gRPC/rust-server/pull/95/files#diff-15e2fc63b62ff238f0f76224e05b1a4641dbeface72dcb2cb7fd1f6f7451241bR36

With #95, the content of `lua/Hooks/DCS-gRPC.lua` is basically what you have in the README in your PR here, just with the addition of loading the config (to get eg the `luaPath`).

Or am I missing something?

justin-lovell: I missed that part.

I'll close this PR because regardless, any improvements made should be done to `main` and it was a significant change made there... making this PR almost moot. 

I'll assess the current workflow introduced #95 and open / suggest improvements on top of that neat rewrite.

PR #91, opened 2021-12-30
split lua files into DCS-gRPC and Hooks

rkusa:
To reduce confusion when developing on DCS-gRPC, I've moved the Lua
files around to map the directory structure once installed as a mod into
DCS more closely.

This means:
- move `lua/lua5.1` to `src/lua5.1` as it is only necessary to build the
  Rust code (which also lives inside of `src/)
- move all other lua files from `lua/*.lua` into `lua/DCS-gRPC`
- except `grpc-hook.lua`, which is moved into `lua/Hooks`

This creates a directory where `DCS-gRPC` and `Hooks` live side-by-side
as they also do once installed as a mod into DCS.


README.md
```diff
@@ -168,7 +168,7 @@ For development, update the previously added line in `DCS World\Scripts\MissionS
 - dofile(lfs.writedir()..[[Scripts\DCS-gRPC\grpc-mission.lua]])
 + GRPC = {
 + 	dllPath = [[C:\Development\DCS-gRPC\rust-server\target\debug\]],
-+ 	luaPath = [[C:\Development\DCS-gRPC\rust-server\lua\]]
++ 	luaPath = [[C:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC]]
```
justin-lovell, line 171: Is this the anticipated suggestion inbound on my PR?

justin-lovell (commented):
Do you know why some files were marked as changes instead of the standard rename? Dodgy line endings from prior commits?

README.md
rkusa: Nope, I am going to suggest to:
- remove the settings inside of the `MissionScripting.lua` completely in favor of the settings file we have now
- get rid of the `Data/dcs-grpc.lua` file (also in favor of the settings file we have now)
- integrate the change you've added to the README in your PR into the project (so having a `lua/Hooks/DCS-grpc.lua` which loads settings and then `lua/DCS-gRPC/dcs-hook.lua`; where the latter is then used for actual logic inside of the hooks environment)

rkusa:
> Do you know why some files were marked as changes instead of the standard rename? Dodgy line endings from prior commits?

Huh, no idea, didn't event notice it. Thanks for pointing it out. I was able to fix it with a manual `git mv`. Thus, also the force-push.

rkusa:
... and another force-push to fix merge conflicts with `main`

rurounijones (changes requested):
Looks good.

Can you update the path in https://github.com/DCS-gRPC/rust-server/blob/main/.gitattributes so that the filthy C of the lua lib doesn't infect our glorious rust & lua statistics.

rurounijones (approved):

PR #93, opened 2022-01-06
Allow forks to CI check before PR

justin-lovell:
Let's improve the Quality of Life for third party contributors by allowing them to anticipate whether or not their PR would satisfy the CI-gates.

This would also free up the maintainers from anticipating issues, and allow the linting checks to be automated.

image

This would not generate a security issue as these builds are pinned to the forked repository. It still requires consent from the maintainers to execute the `pull_request` trigger from third parties. (This PR being a self-verification :) )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: ["*"]
   pull_request:
     branches: [main]
 
--- a/README.md
+++ b/README.md
@@ -221,6 +221,11 @@ return Group.getByName('Aerial-1'):getName()
 
 The REPL is also available in the release and can be run by running `Tools/DCS-gRPC/repl.exe`
 
+### Contributions
+
+This repository is powered by GitHub Actions for the Continuous Integration (CI) services. The same CI checks would be triggered and executed as you push code to your forked repository, and providing early feedback before a maintainer executes a manual execution on the pull request.
+
+
 ### Troublshooting
 
 #### Linker Error 1104


justin-lovell:
And verification. 
image

README.md
```diff
@@ -221,6 +221,11 @@ return Group.getByName('Aerial-1'):getName()
 
 The REPL is also available in the release and can be run by running `Tools/DCS-gRPC/repl.exe`
 
+### Contributions
+
+This repository is powered by GitHub Actions for the Continuous Integration (CI) services. The same CI checks would be triggered and executed as you push code to your forked repository, and providing early feedback before a maintainer executes a manual execution on the pull request.
+
+
```
rkusa, line 228: IMO, since it doesn't need any extra step to set up, I am not entirely sure if it is worth adding this paragraph to the README? Maybe it is something you can expect to just work?

rkusa (approved):
Sounds good to me 👍

README.md
rurounijones: I think it helps set expectations so lets keep it in.

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: ["*"]
   pull_request:
     branches: [main]
```
rkusa, line 7: I wonder if we have to remove the `pull_request` section here, because all checks are now running twice on PRs 🤔

rkusa: Nevermind, I think we should keep `pull_request` as it runs checks on an imaginary merge into `main` (eg: uses  ref `refs/remotes/pull/95/merge` while `push` uses the actual branch `refs/remotes/origin/config-only2`)

PR #94, opened 2022-01-06
send_chat and send_chat_to API

justin-lovell:
* Started the scaffolding for the `NetService`
* `send_chat` implemented
* `send_chat_to` implemented.


protos/dcs/net/v0/net.proto
```diff
@@ -0,0 +1,26 @@
...
+  rpc SendChatTo(SendChatToRequest) returns (SendChatToResponse) {}
+
+  // https://wiki.hoggitworld.com/view/DCS_func_send_chat
+  rpc SendChat(SendChatRequest) returns (SendChatResponse) {}
+}
+
+message SendChatToRequest {
+  string message = 1;
+  uint32 to_player_id = 2;
+  optional uint32 from_player_id = 3;
```
rurounijones, line 16: I am kind of iffy about this, not the fact we added it here, but the whole idea of spoofing another player's ID being part of the API. That just seems like something that is abusable. I know that originally it could only be called in a trusted setting and DCS-gRPC clients should _still_ be trusted but I am still uneasy about it.

Can we think of any situations where we acually could use this?

justin-lovell: I understand your concern of a message being impersonated by another player, and the DCS-gRPC service facilitating that vector. Consequences of emulating social abuse is far more damaging to the community than "system abuse" where cheaters and hackers use to gain an advantage is repairable.

When I was writing that line of code, I was only thinking about the clients/DCS-gRPC/server being threat-free from external malicious code; not considering the severity of the breach.

Given that I cannot think of a good motivation to mimic the `fromPlayerId` parameter (where it supersedes alternative means), I will redact these parameters with a comment explaining the reasons for the redaction. If someone comes up with a good use case, they are able to present their case.

justin-lovell: Done in: https://github.com/DCS-gRPC/rust-server/pull/94/commits/94f8307fa34c4e79a10d359fe57b1863453eaee0

protos/dcs/net/v0/net.proto
```diff
@@ -0,0 +1,29 @@
...
+  uint32 to_player_id = 2;
+}
+
+message SendChatToResponse {}
+
+message SendChatRequest {
+  // the message to send in the chat
+  string message = 1;
+  // which coalition? DCS only supports ALL or NEUTRAL
+  dcs.common.v0.Coalition coalition = 4;
```
rurounijones, line 26: Change `4` to `2`

protos/dcs/net/v0/net.proto
```diff
@@ -0,0 +1,29 @@
...
+
+  // https://wiki.hoggitworld.com/view/DCS_func_send_chat
+  rpc SendChat(SendChatRequest) returns (SendChatResponse) {}
+}
+
+message SendChatToRequest {
+  // the message to send in the chat
+  string message = 1;
+  // the target player of the direct message
+  uint32 to_player_id = 2;
```
rurounijones, line 17: Ignorable Nit: `target_player_id` sounds a bit better to me.

rurounijones (changes requested):

protos/dcs/dcs.proto
```diff
@@ -10,6 +10,7 @@ import "dcs/custom/v0/custom.proto";
 import "dcs/group/v0/group.proto";
 import "dcs/hook/v0/hook.proto";
 import "dcs/mission/v0/mission.proto";
+import "dcs/net/v0/net.proto";
```
rkusa, line 13: I agree that it makes sense to keep close to DCS' structure, which is why we should probably move [`DCS.GetMissionName`](https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/hook/v0/hook.proto#L7) out into its own namespace eventually, too

lua/DCS-gRPC/methods/net.lua
```diff
@@ -0,0 +1,26 @@
...
+
+GRPC.methods.sendChat = function(params)
+    if params.coalition > 1 then
+        return GRPC.errorInvalidArgument("Chat messages can only be sent to all or neutral/spectators")
+    end
+
+    local toAll = true
+    if params.coalition == 1 then
+        toAll = false
+    end
```
rkusa, line 22: Nit (feel free to ignore):
```suggestion
    local toAll = params.coalition != 1
```

rkusa (approved):

rurounijones (approved):

rurounijones (changes requested):
Could you squash all the commits into one and add a git commit message as described in the CONTRIBUTING.md that is being written in https://github.com/DCS-gRPC/rust-server/pull/100

Once that is done I will merge and make a 0.3.0 release with this change included.

PR #95, opened 2022-01-07
use config file as only source for settings

rkusa:
Before this commit it was possible to set settings either:
1) before the `dofile` inside of `MissionScription.lua`,
2) before `GRPC.load()` inside of a mission, and
3) inside of the config file at `Saved Games\DCS\Config\dcs-grpc.lua`.

This commit
a) removes (1),
b) documents the limitations of (2),
c) splits `Hooks/DCS-gRPC.lua` into two parts, and
c) removes all unnecessary `_G.` prefixes on globals (in Lua code; just
   a cleanup as a side-effect of the refactoring).

Reducing the amount of locations that settings can be configured (a)
simplifies the code and reduces the amount things that need to be
tested. More importantly, it removes the need of the temporary
`Data/dcs-grpc.lua` file (which is removed).

With (c) it is now easier to develop the hooks part of DCS-gRPC as it
allows a setup where the actual hook logic (now in
`lua\DCS-gRPC\grpc-hook.lua`) can be loaded from a local repository.
`lua\Hooks\DCS-gRPC.lua` is still there, but it should rarely change
during development as it is only responsible for loading the config and
then executing `lua\DCS-gRPC\grpc-hook.lua` (where the actual hook logic
is implemented).

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,22 @@ For development, update the previously added line in `DCS World\Scripts\MissionS
 + dofile(GRPC.luaPath .. [[grpc-mission.lua]])
 ```
 
+In addition, you will need to replace the hook script at `Scripts\Hooks\DCS-gRPC.lua` to load up your current in development hook script:
+
+```lua
+function init()
+    log.write("[GRPC-Hook-DEV]", log.INFO, "Initializing ...")
+    dofile(GRPC.luaPath .. [[grpc-hook.lua]])
+    log.write("[GRPC-Hook-DEV]", log.INFO, "Initialized ...")
+end
+
+
+local ok, err = pcall(init)
+if not ok then
+  log.write("[GRPC-Hook-DEV]", log.ERROR, "Failed to Initialize: "..tostring(err))
+end
+```
+
 ### Debugging
 
 - Search for `[GRPC]` in the DCS logs
--- a/README.md
+++ b/README.md
@@ -51,68 +51,42 @@ Add the following code to your mission. This will start the DCS-gRPC server. You
 trigger in your .miz file or you can add this code to an existing lua file that your mission may be running.
 
 ```lua
--- Optional: Change settings if desired, e.g.:
-GRPC.debug = true
-
--- Required: Load the gRPC server into the mission
+-- Load the gRPC server into the mission
 GRPC.load()
 ```
 
 ### Settings
 
-The behaviour of the gRPC server can be fine-tuned using various settings that can be set on the `GRPC` global (before the `grpc.lua` is executed). The available settings and their defaults are:
+The behaviour of the gRPC server can be fine-tuned using various settings that can be set inside of `Saved Games\DCS\Config\dcs-grpc.lua` (must be created if it doesn't exist). The available settings and their defaults are:
 
 ```lua
 -- Whether the `Eval` method is enabled or not.
-GRPC.evalEnabled = false
+evalEnabled = false
 
 -- The host the gRPC listens on (use "0.0.0.0" to listen on all IP addresses of the host).
-GRPC.host = '127.0.0.1'
+host = '127.0.0.1'
 
 -- The port to listen on.
-GRPC.port = 50051
+port = 50051
 
 -- Whether debug logging is enabled or not.
-GRPC.debug = false
+debug = false
 
 -- Limit of calls per second that are executed inside of the mission scripting environment.
-GRPC.throughputLimit = 600
+throughputLimit = 600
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -160,18 +134,21 @@ cargo build --features hot-reload
...
-+ GRPC = {
-+ 	dllPath = [[C:\Development\DCS-gRPC\rust-server\target\debug\]],
-+ 	luaPath = [[C:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC]]
-+ }
-+ dofile(GRPC.luaPath .. [[grpc-mission.lua]])
-```
+- update your `MissionScripting.lua` to load `grpc-mission.lua` from your local clone, e.g.:
+  ```diff
+  - dofile(lfs.writedir()..[[Scripts\DCS-gRPC\grpc-mission.lua]])
+  + dofile([[C:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc-mission.lua]])
```
rkusa, line 144: Only drawback of this PR is that you cannot simply do `dofile(GRPC.luaPath .. [[grpc-mission.lua]])` anymore, since `GRPC.luaPath` isn't loaded yet.

lua/Hooks/DCS-gRPC.lua
```diff
@@ -17,91 +33,7 @@ local function load()
...
-      event = {
-        type = "disconnect",
-        id = id,
-        reason = reason + 1, -- Increment for non zero-indexed gRPC enum
-      },
-    })
-  end
-
-  DCS.setUserCallbacks(handler)
+  dofile(GRPC.luaPath .. [[grpc-hook.lua]])
```
rkusa, line 36: Moved the user callbacks into `grpc-hook.lua`

lua/DCS-gRPC/grpc-mission.lua
```diff
@@ -28,21 +28,8 @@ if GRPC.throughputLimit == nil or GRPC.throughputLimit == 0 or not type(GRPC.thr
...
-if file then
-  file:write(
-    "luaPath = [[" .. GRPC.luaPath .. "]]\n"
-    .. "dllPath = [[" .. GRPC.dllPath .. "]]\n"
-    .. "throughputLimit = [[" .. GRPC.throughputLimit .. "]]\n"
-  )
-  file:flush()
-  file:close()
-else
-  env.error("[GRPC] Error writing config: " .. err)
```
rkusa, line 45: No need for `Data\dcs-grpc.lua` anymore

lua/DCS-gRPC/grpc-hook.lua
```diff
@@ -0,0 +1,86 @@
...
+    time = DCS.getModelTime(),
+    event = {
+      type = "disconnect",
+      id = id,
+      reason = reason + 1, -- Increment for non zero-indexed gRPC enum
+    },
+  })
+end
+
+DCS.setUserCallbacks(handler)
```
rkusa, line 86: Mostly extracted from `lua\Hooks\DCS-gRPC.lua`

README.md
justin-lovell: I think this is forgivable for a development workflow. :) (I don't mind because I appreciate that LUA environments are really segregated)

Alternatively, we could make the syntax ugly and do a dirty load on `dofile` load.

Another option is to advise the developer to do a symbolic link on the lua part of the app. https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/mklink

justin-lovell (commented):
> More importantly, it removes the need of the temporary Data/dcs-grpc.lua file (which is removed).

Nice! It would be a lot more consistent.

rurounijones (commented):
Based on questions I have received in the past I think this current README is going to confuse server owners bceause there is no hint that the "prepare" mission and the config file are separate options with no dependencies between them.

I think trying to update the README via this PR will be a bit messy so I think the best thing to do will be to merge this PR then I will create a new one that _only_ updates the README based on behavious I have observed.

justin-lovell:
Time for `CONTRIBUTORS.md`?

rkusa:
> I think trying to update the README via this PR will be a bit messy so I think the best thing to do will be to merge this PR then I will create a new one that _only_ updates the README based on behavious I have observed.

Fine for me 👍

rurounijones (approved):

rkusa:
> Another option is to advise the developer to do a symbolic link on the lua part of the app.

True. Actually something we had in the README before but removed, because we wanted to focus on a setup where it is easy to switch between a dev and non-dev installation.

PR #98, opened 2022-01-09
feat: add missionCommand APIs

b1naryth1ef:
This PR adds support for the mission `missionCommand` API. Due to the structure of this DCS API I've wrapped the multiple functions up into 3 top-level RPCs:

- **AddMissionCommand** wraps `addCommand`, `addCommandForCoalition`, `addCommandForGroup`. Alongside the arguments you provide a string identifier which will be used for the `MissionCommandEvent` to denote which command is being triggered. I considered just providing the DCS name + target (coalition or group) data, however allowing clients to provide a custom identifier seems to support some use cases where names might be similar or duplicated.
- **AddMissionCommandSubMenu** wraps `addSubMenu`, `addSubMenuForCoalition`, `addSubMenuForGroup`. The naming includes `Command` here to make it clear this is part of the `missionCommand` API, and to better-align with the generic `RemoveMissionCommandItem` RPC.
- **RemoveMissionCommandItem** wraps `removeItem`, `removeItemForCoalition`, `removeItemForGroup`. The naming again makes it clear this is part of the `missionCommand` API.

Additionally it adds a new `MissionCommandEvent` which is triggered for any execution of a mission command registered with the `AddMissionCommand` RPC. The only data provided is the unique identifier for the command passed when registering.

## Questions

- I took the liberty of re-exporting the native `event` function under `GRPC` so that we can dispatch events from mission services, but I don't 100% understand whats going on there so its possible this is invalid and will cause some problems with hot-reloading or mission changing.
- I'd argue `missionCommand`'s is actually one of the more clean ED API's, but providing a one-to-one mapping ends up being very messy and verbose. I'm curious what people think, but to me this abstraction seems cleaner. I am also considering splitting up global and coalition/group RPCs as a middle-ground.
- I opted to pass the string group-name instead of the integer group ID. I'd actually prefer to pass group IDs and require users lookup group names themselves, but I think existing parts of the API prefer the string version.


protos/dcs/mission/v0/mission.proto
```diff
@@ -355,6 +372,8 @@ message StreamEventsResponse {
     DisconnectReason reason = 2;
   }
 
+  message MissionCommandEvent { string id = 1; }
```
justin-lovell, line 375: The other day, I spotted this discussion in the archives: https://github.com/DCS-gRPC/rust-server/discussions/67#discussioncomment-1587771

Having a "marker" for the `callback` is a useful establishment over the loose `google.protobuf.Struct` for contract reasons.

Suggestions I would offer:
* Rename `id` to `stateId`.
* Rebuild whether it is a global / coalition / group level `stateId`. This can be enclosed in the parameters made in the `addCommand` step.

protos/dcs/mission/v0/mission.proto
```diff
@@ -434,12 +455,45 @@ message StreamUnitsResponse {
...
+
+message RemoveMissionCommandItemResponse {}
+
+message AddMissionCommandSubMenuRequest {
+  string name = 1;
+  repeated string path = 2;
+  oneof target {
+    bool global = 3;
+    dcs.common.v0.Coalition coalition = 4;
+    string group = 5;
```
justin-lovell, line 495:
```suggestion
    string groupName = 5;
```
I've seen a few API's refer to it as `groupId` or `groupName`.

Yeah, I agree that we should try to stick to `groupId` as much as possible.

justin-lovell:
> I took the liberty of re-exporting the native event function under GRPC so that we can dispatch events from mission services, but I don't 100% understand whats going on there so its possible this is invalid and will cause some problems with hot-reloading or mission changing.

This appears fine to me. Synchronizing an event timeline of when things arrive and processed on the DCS Server to the gRPC Clients would always be a problem if there are multiple streams.

> I'd argue missionCommand's is actually one of the more clean ED API's, but providing a one-to-one mapping ends up being very messy and verbose. I'm curious what people think, but to me this abstraction seems cleaner. I am also considering splitting up global and coalition/group RPCs as a middle-ground.

The preference would to have one gRPC API to a DCS API. I personally floated the idea of "a function overload" but as @rurounijones pointed out, keeping the gRPC interface as simple as possible is by miles easier to debug / reason inter-process issues. The fact we are able to call DCS Mission from external services is a mod-send (if you can spot the pun there :) )

protos/dcs/mission/v0/mission.proto
justin-lovell: I'm changing my suggestion to `groupName` - see the motivation as described here: https://github.com/DCS-gRPC/rust-server/pull/11#issuecomment-903350183

protos/dcs/mission/v0/mission.proto
```diff
@@ -346,7 +362,8 @@ message StreamEventsResponse {
   }
 
   /**
-   * Fired when a player disconnected from the server (not fired for the server's player).
+   * Fired when a player disconnected from the server (not fired for the
+   * server's player).
```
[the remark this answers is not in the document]
b1naryth1ef: @rurounijones how did you originally format the protos? The above is from `clang-format` v10, so I'm not sure if this diff is just drift from people not formatting or whether this is coming from something else.

protos/dcs/mission/v0/mission.proto
```diff
@@ -434,12 +455,45 @@ message StreamUnitsResponse {
...
+  string name = 1;
+  repeated string path = 2;
+  oneof target {
+    bool global = 3;
+    dcs.common.v0.Coalition coalition = 4;
+    string group = 5;
+  }
+}
+
+message AddMissionCommandSubMenuResponse {}
```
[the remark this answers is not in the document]
b1naryth1ef: Do you mean parity with the DCS API function signatures? This value can be trivially derived from the `Request` payload, so I don't think it makes sense to increase the surface area here.

b1naryth1ef:
> The preference would to have one gRPC API to a DCS API.

It'd probably make the most sense then to introduce a `MissionCommands` interface with the directly mapped functions and a separate streaming interface.

protos/dcs/mission/v0/mission.proto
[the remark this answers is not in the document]
rurounijones: We are using https://github.com/yoheimuta/protolint - I also looked at clang and they definitely did things differently. However I don't think everyone is using protolint so is is possible there is drift.

lua/DCS-gRPC/grpc.lua
```diff
@@ -131,6 +131,7 @@ GRPC.errorPermissionDenied = function(msg)
   }
 end
 
+GRPC.event = grpc.event
```
rurounijones, line 134: Could you add a comment here explaining why this line is present. I know you mentioned it in the summary but it would save a future dev a round-trip to the `blame` command and commit history.

lua/DCS-gRPC/methods/mission.lua
```diff
@@ -411,3 +413,59 @@ GRPC.onDcsEvent = function(event)
...
+    time = timer.getTime(),
+    event = {
+      type = "missionCommand",
+      id = callbackId,
+    }
+  })
+  return GRPC.success(nil)
+end
+
+GRPC.methods.addMissionCommand = function(params)
```
rurounijones, line 428: > I'd argue missionCommand's is actually one of the more clean ED API's, but providing a one-to-one mapping ends up being very messy and verbose. I'm curious what people think, but to me this abstraction seems cleaner.

I do not believe it is that messy or verbose and a bit of verbosity in the server-side code is preferable to having to make clients written in different languages dea with extra protocol complexity with `oneofs` and the like.

I prefer we keep everything as simple as possible and split the APIs to, in this case, follow the ED style AddMissionCommandForCoalition etc.

lua/DCS-gRPC/methods/mission.lua
```diff
@@ -411,3 +413,59 @@ GRPC.onDcsEvent = function(event)
     return nil
   end
 end
+
+function missionCommandCallback(callbackId)
```
rurounijones, line 417: Is the intention here being that DCS-gRPC clients will recieve notifications when a radio command (added via thse APIs) has been triggered by a player?

If so then I don't think this is how it works but I will let @rkusa weigh in on this since he is the expert on the streaming side of things.

protos/dcs/mission/v0/mission.proto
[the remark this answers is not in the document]
rurounijones: I would include it in the response. It is easy and if nothing else it allows the requester to optionally do a sanity check and protect against ED doing unexpected ED things like generating a different ID for _reasons_.

protos/dcs/mission/v0/mission.proto
rurounijones: `stateId` sounds wierd and I don't really understand the reasoning for it. I am also not sure what the "marker" is about. Could you add some more clarification around the reasoning for this suggestion?

rurounijones (commented):
Thanks for the contribution!

Could you squash the commits and update the Commit message to include background information on the change, why it is being made, and any answers to potential "Why did you do it that way" questions (such as re-exporting the `GRPC.event` thing). Can you also clarify if this is intended to satisfy part of https://github.com/DCS-gRPC/rust-server/discussions/67 of if it is coincidence?

Commit message format

```
50 character summary

Commit message body. Each line limited to 72'ish characters.
URLs and other strings that cannot be broken up being an exception.
Note the blank line between summary and body.
```

> I took the liberty of re-exporting the native event function under GRPC so that we can dispatch events from mission services, but I don't 100% understand whats going on there so its possible this is invalid and will cause some problems with hot-reloading or mission changing.

Touched on in the review.
> I'd argue missionCommand's is actually one of the more clean ED API's, but providing a one-to-one mapping ends up being very messy and verbose. I'm curious what people think, but to me this abstraction seems cleaner. I am also considering splitting up global and coalition/group RPCs as a middle-ground.

Touched on in a comment in this review

> I opted to pass the string group-name instead of the integer group ID. I'd actually prefer to pass group IDs and require users lookup group names themselves, but I think existing parts of the API prefer the string version.

Yeah. Many ED APIs _only_ work with group name so we standardised on that.

protos/dcs/mission/v0/mission.proto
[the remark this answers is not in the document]
justin-lovell: Consider the following pseudo code snippets.
```
// cascading a path (the suggestion)
var lvl1 = grpcDcs.addSubMenu("First Level");
var lvl2 = grpcDcs.addSubMenu("Second Level", lvl1.path);
var menu = grpcDcs.addCommand("command", lvl2.path);
```
vs
```
// deriving the path
var path = [];
grpcDcs.addSubMenu("First Level", path);
path.push("First Level"); // or even extract the string as a local var...
const secondLevel = "Second Level";
grpcDcs.addSubMenu(secondLevel, [ "Or Maybe Prone to typo" ]);
path.push(secondLevel);
grpcDcs.addCommand("command", path);
```

So yes, I can see how it can be derived; it's not a show stopper. Both of us could agree if there is a minor change to our code, we make the client code feel neater when the effort is of similar nature.

https://github.com/DCS-gRPC/rust-server/blob/2db3b3bf4b1be0f7628a8171356198770d59dd89/src/rpc/atmosphere.rs#L8-L18
https://github.com/DCS-gRPC/rust-server/blob/15f806463cc7df23ac9c73d67c90fa0a8dbc1d76/lua/DCS-gRPC/methods/atmosphere.lua#L6-L10

protos/dcs/mission/v0/mission.proto
justin-lovell: Consider the client code has something like the following pseudo:
```
var commandTable = new Dictionary<string, Action>();
// key is state ID
var enrolledDateTime = (new DateTime()).ToString();
var commandStateId = Guid.New().ToString();
commandTable.Add(commandStateId, () => Console.WriteLine("Hello " + enrolledDateTime));

// ...
grpcDcs.AddMissionCommand({StateId = commandStateId});

// ...
grpcDcs.OnMissionEvent(theEvent => {
    var theCommand = theEvent as MissionCommandEvent;
    if (theCommand != null) {
       Action restoredState = commandTable[theCommand.stateId];
       restoredState();
    }
```
The main idea is that it makes the client application responsible in restoring the state for the event, rather than relying on the prior state being passed around from the gRPC->DCS->gRPC layer. It removes the need to have an open struct as mentioned in the discussion I linked to, thus eliminate the dumping group that a mission command could become.

protos/dcs/mission/v0/mission.proto
```diff
@@ -434,12 +455,47 @@ message StreamUnitsResponse {
...
 
 message GetScenarioCurrentTimeRequest {}
 
-message GetScenarioCurrentTimeResponse {
-  string datetime = 1;
+message GetScenarioCurrentTimeResponse { string datetime = 1; }
+
+message AddMissionCommandRequest {
+  string id = 1;
+  string name = 2;
```
rurounijones, line 466: I would rename these fields to reduce confusion (See the #developers channel on the discord to see what I mean).

id -> "command_name"
name -> "menu_text"

Also. Add documentation to these fields along the lines of :

* https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/unit/v0/unit.proto#L70
* https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/common/v0/common.proto#L184

protos/dcs/mission/v0/mission.proto
rurounijones: Talked in Discord. My suggestion here is https://github.com/DCS-gRPC/rust-server/pull/98/files?diff=split&w=1#r780785475 in the request and in the CommandEvent we just use `name` or `command_name` instead of `id`

protos/dcs/mission/v0/mission.proto
rkusa: When I am going to use this API, I'll probably add a JSON encoded payload as the `id`/`command_name` as most of my use cases have both a command name and one or more arguments. So a more general field time like simply `command` might be more appropriate - emphasising that you can add whatever string you like?

lua/DCS-gRPC/methods/mission.lua
rkusa: I would have added this as a custom event, too. So it is good that way IMO

protos/dcs/mission/v0/mission.proto
rkusa: I don't like the idea of having to keep track which id corresponds to which logic in my client. So I'd personally either use the field to send JSON encoded data (thus the suggestion to rename it to `command` in another comment), or'd suggest to use something like:

```proto3
string command = 1;
repeated google.protobuf.Struct args = 2;
```

I don't have a strong opinion on whether to keep the single string field, or the suggestion above, both would work for my use-cases equally fine. But I am on-board of using another name for the field.

rkusa (commented):
Thanks a lot for the PR! I hope you don't mind all the comments we left. It's just a result of this API being a more delicate one. I am looking forward to have this included in DCS-gRPC (need it myself soon too).

protos/dcs/mission/v0/mission.proto
justin-lovell: Another suggestion: ‘key’ for command name? (I would also be encoding a json payload myself)

rurounijones:
Before we make any more suggestions can we confirm what the scope of this API is supposed to be. My impression is that we have a mix of `Send an event that a command with the ID of X has been triggered via a radio menu`  and `Send an event with an arbitrary payload`

lua/DCS-gRPC/methods/mission.lua
b1naryth1ef: ah yeah no need for the `GRPC.success` here

justin-lovell:
I definitely agree that the scope of this PR should be for the following:

> "Send an event that a command with the ID of X has been triggered via a radio menu"

PS: I did link to the #67 discussion due to the last two messages regarding the radio menu "context"/"state"/"command id"/"command key" (which did diverge from the original message in that thread) but had some context of the API design of the radio menu. In no way was I suggesting this PR addresses the _original_ message of that discussion.

protos/dcs/mission/v0/mission.proto
rurounijones: Before we discuss this further. See https://github.com/DCS-gRPC/rust-server/pull/98#issuecomment-1008441891

rurounijones:
@b1naryth1ef  What do you think about replacing a "command ID" style call with one where we can pass in an arbitrary struct as described in https://github.com/DCS-gRPC/rust-server/pull/98#discussion_r780790788 ?

PR #103, opened 2022-01-13
CI: Cache build improves build time from 5mins to 1m30s-2m10s

justin-lovell:
I got annoyed by the build times on the CI. 

So here is a fix.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,16 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
       - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
         with:


rurounijones:
Would it be more future-proof to use this rust specific cache action? https://github.com/marketplace/actions/rust-cache

justin-lovell:
I did come across that package - and it should work too. The only advantage it has over the builtin cache is the following:

> Additionally, the action automatically works around cargo#8603 / actions/cache#403 which would otherwise corrupt the cache _on macOS builds._

Since we aren't targeting that platform, it's kind of a moot point.

The primary motivation for using the built-in cache action is that no needed to review the contents of the action in depth. I personally try to avoid absorbing dependencies unless there is a distinct advantage. Otherwise it becomes the `npm`-wild-west :) 

Having said that, I am not adverse to that option either for this utility function.

Let me know what your preference would be -- I think the cache will help a lot in the feedback loop.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,16 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
```
rkusa, line 28: The [cargo example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) has slightly different entries, I am wondering whether we should go with them instead? It is probably the same, as the example only narrows it down to subdirectories:

```suggestion
            ~/.cargo/bin/
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            target/
```

rkusa (commented):
I am in favour of directly using the official `actions/cache` rather than an action (dependency) from a random user from the marketplace. I'd consider copy-and-paste easy enough here to not pull in another dependency.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,16 @@ jobs:
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+            ${{ runner.os }}-cargo
```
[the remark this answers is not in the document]
justin-lovell: The indent on line 30 triggers a "continue as if on same line". They call it "Multi-line Flow Scalars"

.github/workflows/ci.yml
justin-lovell: I didn't find that example. Great find. 

I will align to that example if @rurounijones agrees to keep it simplistic on the dependency chain.

.github/workflows/ci.yml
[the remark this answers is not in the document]
rkusa: Gotcha, thanks for the clarification. If both is basically one line, then I think `${{ runner.os }}-cargo` is duplicated and one can be removed?

.github/workflows/ci.yml
rurounijones: Yeah, lets just copy/paste and not add the dependency

.github/workflows/ci.yml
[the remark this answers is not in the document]
justin-lovell: Yeah, give we are not going for multi-target ... it does "simplify" the line at the expense of being "comprehensive. "

.github/workflows/ci.yml
justin-lovell: Sorted and pushed

PR #106, opened 2022-01-15
Force Player Slot Selection

justin-lovell:
Implementation of the `force_player_slot` function. There is only
community knowledge on how this function operates; it is officially
listed but not publicly documented.

### DIFF
--- a/lua/DCS-gRPC/methods/net.lua
+++ b/lua/DCS-gRPC/methods/net.lua
@@ -41,3 +41,14 @@ GRPC.methods.getPlayers = function()
 
     return GRPC.success({players = players})
 end
+
+GRPC.methods.forcePlayerSlot = function(params)
+    if params.coalition == 0 then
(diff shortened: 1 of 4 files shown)


rurounijones (approved):

lua/DCS-gRPC/methods/net.lua
```diff
@@ -41,3 +41,14 @@ GRPC.methods.getPlayers = function()
 
     return GRPC.success({players = players})
 end
+
+GRPC.methods.forcePlayerSlot = function(params)
```
rurounijones, line 45: Is there a way to force back to spectator via this API? If so, what is the incantation and can we document it.

justin-lovell: Yep! We can force people into very specific Blue/Red slots.

justin-lovell:
I've placed some pseudo example call in the documentation.

PR #109, opened 2022-01-15
Emit when a player changes to a slot

justin-lovell:
The DCS Control API should emit a special event to the gRPC
clients that a player has changed a slot.

```
grpcurl.exe -plaintext -import-path ./protos -proto ./protos/dcs/dcs.proto -d '{}' 127.0.0.1:50051 dcs.mission.v0.MissionService/StreamEvents
{                      
  "time": 26.399,      
  "playerChangeSlot": {
    "playerId": 2,
    "coalition": "COALITION_BLUE",
    "slotId": "817"
  }
}
{
  "time": 33.158,
  "playerChangeSlot": {
    "playerId": 2,
    "coalition": "COALITION_BLUE",
    "slotId": "816"
  }
}
{
  "time": 33.158,
  "playerLeaveUnit": {
    "initiator": {

    }
  }
}

```

### DIFF
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -310,6 +310,16 @@ message StreamEventsResponse {
     // what was typed
     string message = 2;
   }
+
+  // fired when the player changes across to a slot
+  message PlayerChangeSlotEvent {
+    // The player's id in the current server session.
+    uint32 player_id = 1;
+    // The slot's coalition
+    dcs.common.v0.Coalition coalition = 2;
+    // The slot's identifier
+    string slotId = 3;
+  }
   
   /**
    * Fired when a player connected to the server.
@@ -401,6 +411,7 @@ message StreamEventsResponse {
     ConnectEvent connect = 8192;
     DisconnectEvent disconnect = 8193;
     PlayerSendChatEvent player_send_chat = 8194;
+    PlayerChangeSlotEvent player_change_slot = 8195;
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `PlayerChangeSlotEvent` emitted when player changes slot
+
 ### Fixed
 - Corrected `proto` files from camel-casing to snake-casing; not a runtime breaking change but some code generators
   may generate different casing by convention, creating a compiler only issue.
--- a/lua/DCS-gRPC/grpc-hook.lua
+++ b/lua/DCS-gRPC/grpc-hook.lua
@@ -83,4 +83,18 @@ function handler.onPlayerDisconnect(id, reason)
   })
 end
 
+function handler.onPlayerChangeSlot(playerId)
+  local playerInfo = net.get_player_info(playerId)
+
+  grpc.event({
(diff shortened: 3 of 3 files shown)


protos/dcs/mission/v0/mission.proto
```diff
@@ -310,6 +310,16 @@ message StreamEventsResponse {
     // what was typed
     string message = 2;
   }
+
+  // fired when the player changes across to a slot
+  message PlayerChangeSlotEvent {
```
rurounijones, line 315: The handler method is TRYChangeSlot so should this event naming reflect that? is there a handler.OnPlayerSuccessChangeSlot or something that we can also emit an event for?

justin-lovell: Although we are intercepting the DCS's `try...` function, there is no equivalent "did" change slot; it's very similar motivation to how we intercept the `try...chat` message -- the caller expects an synchronous response (pure function) which would not be possible with the async trip the rust-server would need to make to the connected clients.

If no return value is made by the handler - as per the implementation - other hooks such as `slmod` would then execute. We just intercept, pass it up the gRPC clients.

If the gRPC determines it should not be allowed, they are welcome to execute the appropriate `ForcePlayerSlot`.

justin-lovell: Example use case .... https://github.com/ciribob/DCS-SimpleSlotBlock
My extended case will be around a few integration points (e.g. authentication, match organizing, long running missions)

rurounijones: Right, but the event "PlayerChangeSlotEvent", at least to me, implies they _did_ change slot. `PlayerTryChangeSlotEvent` seems a more accurate event name.

justin-lovell: > `PlayerChangeSlotEvent`, at least to me, implies they did change slot. 

I agree with this as the premise "if it is in the past."

Based on the above proposed behavior, from the _gRPC clients' perspective (and timeline)_, it did occur by the time they read the streamed event.

Internally, the following _has_ to occur due to the synchronous behavior of `onPlayerTryChangeSlot`:
1. Read the message - make a decision. 
2. Since it cannot determine an outcome, it does not return a value (cannot make a decision).
3. It would enqueue the event that a player has selected a slot (by implication). This is the part which is in the past tense. 

When the gRPC Client does read the streamed events, the client has a good chance they have been accepted into the slot (assuming no other hook has evicted the person, or they person hasn't changed slots).

So therefore, when the client does read the event, it has occurred in the past. At this point, using a `ForcePlayerSlot` could force a player into a certain workflow (e.g. revoke/eject the player's slot selection).

rurounijones: > I agree with this as the premise "if it is in the past."

> Based on the above proposed behavior, from the gRPC clients' perspective (and timeline), it did occur by the time they read the streamed event.

The client has definitely _tried_ to change slot by the time that the client receives the event but there is no guarantee that it succeeded _after_ the client has read the event since anything else in the handler chain could have denied the change.

This is why having the absolute "The player changed slot" is misleading I think. At the point that the DCS-gRPC client receives the event we have no idea if it succeeded or not. That is why I believe we should rename the event to highlight the fact that the player _tried_ to change slot at the point we receive the message but we have no idea if it was successful.

justin-lovell: Used different hook point.

PR #111, opened 2022-01-15
CI: Lint Check of proto files

justin-lovell:
During the CI build, ensure that the files meet the styleguide.

Some of the errors were resolved in other PR's created.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,3 +68,20 @@ jobs:
       - name: Lint
         run: luacheck ./lua
 
+  proto:
+    name: Proto
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install protolint
+        run: |
+          mkdir -p "$HOME/.local/bin"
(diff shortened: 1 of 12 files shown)


justin-lovell:
Well, it works.
image

rurounijones:
I am not in favour of doing this _yet_. It forces decisions that do not need to be made yet while we still figure things out (the linked stuff about enums, waypoints etc. etc. being a good example).

justin-lovell:
> not need to be forced early just to satisfy a linter

Almost every PR has a mention on how the `protobuf` definitions are not meeting the styleguide, which is what the linter is supposed to enforce.

More importantly -- most of the specific items which the linter is complaining about does not have an implementation / behavior. It should be encouraged that implementation should be followed up on those API's because it has demonstrated that it is expensive to think about whilst it doesn't have an influence on behavior.

justin-lovell:
PR is ready for review.

rurounijones (approved):

Issue #20, opened 2021-08-23
Share unit-stream state between concurrent streams


rkusa:
See https://github.com/DCS-gRPC/rust-server/pull/11#discussion_r679441709

PR #112, opened 2022-01-15
Fixing `GroupCategory` enumeration

justin-lovell:
Motivated by work in progress with PR #111. Communicating
breaking changes and configuring to change log and updating
behavior


justin-lovell:
This would take a while to fully implement because the implementation here is incorrect -- previously it would have only gotten the airplanes only, and not the other unit types.

https://github.com/DCS-gRPC/rust-server/blob/3f4468807b37014140ebb10539d95b649ad6eb55/src/stream.rs#L49-L52

I'm considering we change the `StreamUnitsRequest` to include the `GroupCategory` with the motivation of that the different units move at different speed, and hence it would result in a different poll rate requirement.

src/rpc/mission.rs
```diff
@@ -44,11 +44,19 @@ impl MissionService for MissionRpc {
         &self,
         request: Request<mission::v0::StreamUnitsRequest>,
     ) -> Result<Response<Self::StreamUnitsStream>, Status> {
+        let request_message = request.into_inner();
```
rkusa, line 47: Nit: Might be my personal style, but I think I also see it a lot in other Rust projects. If you don't plan on using the old value of `request` any longer, you can simply re-use it:

```suggestion
        let request = request.into_inner();
```

src/stream.rs
```diff
@@ -26,7 +26,6 @@ pub async fn stream_units(
     rpc: MissionRpc,
     tx: Sender<Result<Update, Status>>,
 ) -> Result<(), Error> {
-    // initialize the state for the current units stream instance
```
rkusa, line 29: Nit: why remove the comment, is it that useless? 😄

src/rpc/mission.rs
```diff
@@ -44,11 +44,19 @@ impl MissionService for MissionRpc {
...
+
+        if request_message.category == 0 {
+            return Err(Status::out_of_range(
+                "Specify the category of the unit to stream",
+            ));
+        }
```
rkusa, line 53: Why not keep support for allowing a stream that includes all unit types - I'd assume not everyone wants to stream different unit types at different update rates?

rurounijones: +1 The clients I am writing don't need this granularity.

Question for @rkusa Given https://github.com/DCS-gRPC/rust-server/issues/20 is still open. Does having potentially many streams active for each client (vs the one-per-client now), like this feature opens up, have any performance implications we need to actually worry about.

rkusa:
If we add this filter, we'd also have to filter the units added in the following line (where it listens for birth events):

https://github.com/DCS-gRPC/rust-server/blob/main/src/stream.rs#L138

src/rpc/mission.rs
rkusa: @rurounijones assuming that the streams observe different units, there is almost no additional work done compared to observing all in one stream.

src/rpc/mission.rs
justin-lovell: I didn't realize Rust would allow a new variable to overshadow an old one! Cool, I have adjusted it locally.

src/stream.rs
justin-lovell: My bad. I had a few lines commented and then just deleted all the comments without realising I took something extra with it.

src/rpc/mission.rs
justin-lovell: I've now implemented `UNSPECIFIED` would monitor all units. If someone desires to have different polling variations, they are welcome to join the observation streams on the client app.

src/stream.rs
```diff
@@ -125,7 +125,7 @@ struct Context {
 }
 
 /// Update the given [State] based on the given [Event].
-async fn handle_event(state: &mut State, event: Event) -> Result<(), Error> {
+async fn handle_event(state: &mut State, event: Event, category: i32) -> Result<(), Error> {
```
justin-lovell, line 128: > If we add this filter, we'd also have to filter the units added in the following line (where it listens for birth events):

Cheers @rkusa for spotting that. Without importing another crate, is the implementation of passing an `i32` instead of `GroupCategory` the best I could do? I tried searching for a few `enum.parse` like functionality in Rust but turning out empty handed.

rkusa: Thanks for taking care of it! You can do `GroupCategory::from_i32(cat)`, it returns a `Option<GroupCategory>`. If you ever want to look at the Rust docs of the code generated from the proto files, you can run the following command (I did the same just now to find this method):


```bash
cargo doc --no-deps -p dcs-grpc-stubs --open
```

rurounijones (dismissed):

rkusa (dismissed):
👍

src/stream.rs
justin-lovell: Ah! Neat trick with generating the documentation with the macros. I was just inferring by convention.

rurounijones:
@justin-lovell It looks, to me, that there are 3 commits in this PR and the latter 2 might accidentally have not been squashed.

* https://github.com/DCS-gRPC/rust-server/pull/112/commits/b85ff5c536c39e31b1e56065bc9da3d345a6d17f
* https://github.com/DCS-gRPC/rust-server/pull/112/commits/a2a97a884aa20dd7289b80e3920d588507c53713

Could you check on your end and squash them if they are supposed to be; or change the commit messages (They are currently identical) if you think they should be separate commits.

rurounijones (changes requested):
@justin-lovell It looks, to me that there are 3 commits in this PR and the latter 2 might accidentally have not been squashed together.

* https://github.com/DCS-gRPC/rust-server/pull/112/commits/b85ff5c536c39e31b1e56065bc9da3d345a6d17f
* https://github.com/DCS-gRPC/rust-server/pull/112/commits/a2a97a884aa20dd7289b80e3920d588507c53713

Could you check on your end and squash them if they are supposed to be; or change the commit messages (They are currently identical) to be more specific to the commit goal if you think they should be separate commits.

justin-lovell:
@rurounijones -- yeah, the last one was my bad. I must have forgotten the amend. 

Having the two commits is recording good history. One was for the linting, and the second one is to align `StreamUnits` up to standard.

rurounijones (approved):

Issue #76, opened 2021-11-12
Group Spawning


rurounijones:
This issue is the master issue for getting group spawning working and will probably have more things added to it over time as we realise what needs to be done:

# Ground Groups
- [ ] Waypoints
- [ ] Tasks

# Helicopter Groups
- [ ] Group
- [ ] Units
- [ ] Waypoints
- [ ] Tasks
- [ ] Fuel/Guns/countermeasures etc.
- [ ] Pylon setup

# Plane Groups
- [ ] Group
- [ ] Units
- [ ] Waypoints
- [ ] Tasks
- [ ] Fuel/Guns/countermeasures etc.
- [ ] Pylon setup

# Sea Groups
- [ ] Group
- [ ] Units
- [ ] Waypoints
- [ ] Tasks

rurounijones:
We can currently spawn a static SAM site (See https://github.com/DCS-gRPC/rust-server/pull/64#issuecomment-965349071 ) however I am not checking anything in the ground group until I have had more time to do testing.

rurounijones:
@Wizxrd Suggested that we pass in an abstraction for the waypoint type and then have the lua fill in the rest: i.e.

```lua
actions = {
  turningPoint      = { name=_('Turning point'),            type='Turning Point',     action='Turning Point' },
  flyOverPoint      = { name=_('Fly over point'),           type='Turning Point',     action='Fly Over Point' },
  finPoint          = { name=_('Fin point (N/A)'),          type='Fin Point',         action='Fin Point' },
  takeoffRunway  	= { name=_('Takeoff from runway')	  , type='TakeOff',           action='From Runway' },
  takeoffParking	= { name=_('Takeoff from parking')	  , type='TakeOffParking',    action='From Parking Area' },
  takeoffParkingHot = { name=_('Takeoff from parking hot'), type='TakeOffParkingHot', action='From Parking Area Hot' },
  LandingReFuAr 	= { name=_('LandingReFuAr')			  , type='LandingReFuAr',     action='LandingReFuAr' },
  
  takeoffGround	    = { name=_('Takeoff from ground')	   ,type='TakeOffGround',     action='From Ground Area' },
  takeoffGroundHot  = { name=_('Takeoff from ground hot'),  type='TakeOffGroundHot',  action='From Ground Area Hot' },
  
  landing 		    = { name=_('Landing'),                  type='Land',              action='Landing' },
[...]

PR #113, opened 2022-01-15
Linting: implemented suggestions for coalition.proto

justin-lovell:
Primary motivation to be a commit is due to CHANGELOG.md update
and for a discussion on the API itself being affected.

Work motivated by corrected #111 but separated PR as there are some discussion points
on how to resolve some design issues. Comment inline the PR below.

### DIFF
--- a/protos/dcs/coalition/v0/coalition.proto
+++ b/protos/dcs/coalition/v0/coalition.proto
@@ -43,7 +43,7 @@ message AddGroupRequest {
     bool late_activation = 3;
     string name = 4;
     dcs.common.v0.Position position = 5;
-    repeated Point route = 6;
+    repeated Point waypoints = 6;
     uint32 start_time = 7;
     string task = 8;
     bool task_selected = 9;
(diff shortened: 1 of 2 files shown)


protos/dcs/coalition/v0/coalition.proto
```diff
@@ -77,25 +77,26 @@ message AddGroupRequest {
...
+    PointType type = 3;
     string action = 4;
     string form = 5;
     double speed = 6;
   }
 
   enum Skill {
-    SKILL_RANDOM = 0;
+    // unspecified skill would use DCS random skill
+    SKILL_UNSPECIFIED = 0;
```
justin-lovell, line 99: I think this is a practical middle ground -- if unspecified, you get the default which is random in DCS MSE.

protos/dcs/coalition/v0/coalition.proto
```diff
@@ -77,25 +77,26 @@ message AddGroupRequest {
...
-      TURNING_POINT = 3;
-      TAKEOFF_PARKING_HOT = 4;
-      LAND = 5;
+    enum PointType {
+      POINT_TYPE_UNSPECIFIED = 0;
+      POINT_TYPE_TAKEOFF = 1;
+      POINT_TYPE_TAKEOFF_PARKING = 2;
+      POINT_TYPE_TURNING_POINT = 3;
+      POINT_TYPE_TAKEOFF_PARKING_HOT = 4;
+      POINT_TYPE_LAND = 5;
```
justin-lovell, line 86: This only applies to planes. I don't think this should be part of the `Point` type. 

We should uplift the starting position and configuration to the various `xxxxGroupTemplate` and append waypoints.

protos/dcs/coalition/v0/coalition.proto
rurounijones: I don't like the "Unspecified means something unexpected". We should increment (and throw error on 0) or leave 0 as RANDOM and have the linter ignore using a comment.

justin-lovell: I've created an exclusion on the rule for this.

justin-lovell (commented):
For this PR, I feel that this is not useful to my time. I personally would rewrite the entire `AddGroupRequest` and friends ... most of the members don't comply with the linting, and most importantly, the **transgressors of the linting issues has no implementation**!

Let us get the linting under order (because that is a common call out on every PR), or just eliminate the members which is not implemented.

protos/dcs/coalition/v0/coalition.proto
rurounijones: See https://github.com/DCS-gRPC/rust-server/issues/76#issuecomment-973463493 for ideas around this.

protos/dcs/coalition/v0/coalition.proto
```diff
@@ -43,7 +43,7 @@ message AddGroupRequest {
     bool late_activation = 3;
     string name = 4;
     dcs.common.v0.Position position = 5;
-    repeated Point route = 6;
+    repeated Point waypoints = 6;
```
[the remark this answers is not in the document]
rurounijones: This isn't renamed due to a linting warning is it? If not can we remove from this PR so that it is limited to linting fixes only.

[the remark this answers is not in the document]
justin-lovell: > This isn't renamed due to a linting warning is it?

The linter is throwing an error because "route" is not a plural.

If `waypoints` is not a suitable substitute, I would recommend the alternatives of:
* Support `route.points` where `route` will only have one property called `points` (the inspiration of the waypoints)
* Alternatively, disable the linter on this line -- subject to the review as we are developing this API.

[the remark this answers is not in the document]
rurounijones: Thanks for the clarification. Lets leave this as is for now then. I think having a `Route` with `Points` as you suggest may be the way to go in the future but we can deal with that when the time comes.

rurounijones (approved):

PR #118, opened 2022-01-29
MarkToCoalition - Fix coalition misalignment

justin-lovell:
Off by one error :)


lua/DCS-gRPC/methods/trigger.lua
```diff
@@ -72,7 +72,8 @@ GRPC.methods.markToCoalition = function(params)
   local point = coord.LLtoLO(params.position.lat, params.position.lon, params.position.alt)
   local idx = getMarkId()
 
-  trigger.action.markToCoalition(idx, params.text, point, params.coalition, params.readOnly, params.message)
+  local coalition = params.coalition - 1 -- Decrement for non zero-indexed gRPC enum
+  trigger.action.markToCoalition(idx, params.text, point, coalition - 1, params.readOnly, params.message)
```
rurounijones, line 76: You have `coalition -1` twice. Once when setting the `coalition` variable on line 75 and once again in the method on line 76.

I think you need to get rid of one of the decrements.

rurounijones (approved):

PR #119, opened 2022-01-29
Add C# headers

rurounijones:
Add headers to the proto files that specify the namespace the generated
C# files will live under. This is part of work to create a simple C#
nuget package so clients to not have to re-invent the wheel in their own
appications.

There are already `Dcs.*` nuget packages so to avoid confusion and
conflict I am putting these under `RurouniJones` as a unique namespace
since we don't have a company or any name that is particularly unique
and I will be publishing other C# DCS related packages so it is as good
a place as any.

-------------

Notes.
I am creating a C# package that I will publish to nuget. I will create a repo in this project called `csharp-client` and I have created a [DCS-gRPC](https://www.nuget.org/organization/DCS-gRPC) organisation on nuget that we can add people to.

### DIFF
--- a/protos/dcs/atmosphere/v0/atmosphere.proto
+++ b/protos/dcs/atmosphere/v0/atmosphere.proto
@@ -1,6 +1,8 @@
 syntax = "proto3";
 package dcs.atmosphere.v0;
 import "dcs/common/v0/common.proto";
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Atmosphere";
+
 
 // https://wiki.hoggitworld.com/view/DCS_singleton_atmosphere
 service AtmosphereService {
--- a/protos/dcs/coalition/v0/coalition.proto
+++ b/protos/dcs/coalition/v0/coalition.proto
@@ -1,6 +1,7 @@
 syntax = "proto3";
 package dcs.coalition.v0;
 import "dcs/common/v0/common.proto";
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Coalition";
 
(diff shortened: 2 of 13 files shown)


rkusa:
I am not against the addition, just out of curiosity, would `--csharp_opt=base_namespace=RurouniJones` for the `protoc ` command an option, too (docs: https://developers.google.com/protocol-buffers/docs/reference/csharp-generated#compiler_options)? This way the protos itself wouldn't have to concern themselves with the C# code generation, and we wouldn't have to remember to add `csharp_namespace` for each new proto file.

rurounijones:
I have had a hell of a time trying to get the protoc stuff working in the past with options. I will try again just to be sure.

rurounijones:
So. Based on what I am seeing elsewhere it seems that the basename thing is... convoluted when it comes to using it with `Grpc.Tools` (The standard dependency used for this stuff in C#).

* https://github.com/grpc/grpc/issues/17672
* https://github.com/grpc/grpc/issues/28663

The microsoft documentation itself uses the namespace in the protos files: https://docs.microsoft.com/en-us/aspnet/core/grpc/basics?view=aspnetcore-6.0

In the interests of possibly not going down _another_ rabbit hole of compatibility I would like to merge this PR as is and then investigate later to see if we can get rid of it and instead have the basename gone in the client package only.

I believe we will need to be adding Golang specific headers to these files anyway so even if we get rid of the C# stuff we will need to remember to add the golang stuff.

justin-lovell:
It might also denote to others that the package managed versions are first party to this repository

rkusa (approved):

PR #121, opened 2022-02-04
Add F10 radio mission commands

rurounijones:
Add MissionCommand API support

    Add support for the `AddMissionCommand`, `AddMissionCommandSubMenu` and
    `RemoveMissionCommandItem` APIs along with their Coalition and Group
    variants which allow us to create F10 radio command menu entries and
    have them emit events to DCS-gRPC clients.

    Additionally it adds new CommandEvents which are triggered for
    any execution of a command via the F10 radio menu by the players.

### DIFF
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -1,8 +1,14 @@
 syntax = "proto3";
 package dcs.mission.v0;
 import "dcs/common/v0/common.proto";
+import "google/protobuf/struct.proto";
 option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Mission";
 option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/mission";
+
+// the "path" field in mission command requests and responses is a repeated
+// string however "paths" doesn't make sense. therefore we will disable
+// the linter pluralization checks for this file.
+// protolint:disable REPEATED_FIELD_NAMES_PLURALIZED
 
 // Contains the streaming APIs that streaming information out of the DCS server.
 service MissionService {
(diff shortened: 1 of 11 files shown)


rurounijones:
The API has been implemented for the global mission command adding and removing and it has been tested and confirmed that the events are emitted. There are some OpenQuestions

# The "name" field

While I was testing the commands the I found the "name" field to be really confusing as to what it meant. I think a better name for this field would be something along the lines of `menuText`, `displayText` or something along those lines

# Coalition & Group specific events

For the Coalition and Group levels we have two options with regards to fields and the events emitted.

1. Do nothing special. The emitted event will be identical to the event emitted for the global Mission command when it is triggered, any information such as the initiator will have to be specified in the Struct.

```
  local event = {
    type = "missionCommand",
    details = details,
  }

  GRPC.event({
    time = timer.getTime(),
    event = event
  })
```

2. Emit different event types and hardcoded fields. For example if we added a command for a group then instead of emitting `missionCommand` event we would emit a `groupMissionCommand` event which has the groupName as a specific field. For example:

```
  local event = {
    type = "groupMissionCommand",
    groupName = "group1,"
    details = details,
  }

  GRPC.event({
    time = timer.getTime(),
    event = event
  })
```

I favour the second option since the more we can remove from the `Struct` the better. The cons are having three separate event types which I don't consider a problem.

Opinions? Once we have answered these two questions I will implement the Coalition and Group level APIs.

protos/dcs/mission/v0/mission.proto
```diff
@@ -420,7 +445,7 @@ message StreamEventsResponse {
     ConnectEvent connect = 8192;
     DisconnectEvent disconnect = 8193;
     PlayerSendChatEvent player_send_chat = 8194;
-    PlayerChangeSlotEvent player_change_slot = 8195;
+    MissionCommandEvent mission_command = 8195;
```
rkusa, line 448: Removing `player_change_slot` was probably not intended?

protos/dcs/mission/v0/mission.proto
```diff
@@ -471,3 +496,29 @@ message GetScenarioCurrentTimeRequest {
 message GetScenarioCurrentTimeResponse {
   string datetime = 1;
 }
+
+message AddMissionCommandRequest {
+  string name = 1;
+  repeated string path = 2;
```
rkusa, line 502: How'd you feel about removing `AddMissionCommandSubMenuRequest` in favour of allowing an empty path for root elements?

rkusa (commented):
Great work, thanks for taking over the PR!

> # The "name" field
> While I was testing the commands the I found the "name" field to be really confusing as to what it meant (and its relationship to the path). I think a better name for this field would be something along the lines of `menuText`, `displayText` or something along those lines.

I personally find `menuText`, `displayText` more confusing than `name` - very subjective though.

> # Coalition & Group specific events
> [...]
> I favour the second option since the more we can remove from the `Struct` the better. The cons are having three separate event types which I don't consider a problem.

Of those two options, I'd prefer the second, too. How about a third: make the group optional and re-use the same event for both?

rurounijones:
> Of those two options, I'd prefer the second, too. How about a third: make the group optional and re-use the same event for both?

Re-using event types with optional fields (`groupName` and `coalitionId` in this case) sucks for statically typed languages when writing switch/case-based event handling code since you cannot just base your logic on the event type but then need an inner set of logic checking for whether certain fields are present and populated. Much nice to just be able to base it on the event type itself.

protos/dcs/mission/v0/mission.proto
rurounijones: Probably a rebase cockup, I shall fix.

rurounijones:
Went with the second option of havine different events

justin-lovell (approved):
Everything makes sense to me -- I would have motivated the same direction as per the decisions made in the PR.

> Re-using event types with optional fields (groupName and coalitionId in this case) sucks when writing switch/case-based event handling code

If we did have a broad event, I would have used a pattern matching approach if I needed to figure out which "level" instead of heavy switch statements. IMHO, I would have been deriving the context from the `details` alone.

rurounijones:
This will have to be a discussion for later if we go for a broad general purpose event but not all client languages have structural pattern matching but everyone can do switch statements.

PR #122, opened 2022-02-06
Add exporters for Static and Scenery objects

rurounijones:
Fill out the fields for the exporters for `Static` and `Scenery` objects based on the methods in.

* https://wiki.hoggitworld.com/view/DCS_Class_Static_Object
* https://wiki.hoggitworld.com/view/DCS_Class_Scenery_Object

### DIFF
--- a/protos/dcs/common/v0/common.proto
+++ b/protos/dcs/common/v0/common.proto
@@ -280,6 +280,12 @@ message Static {
  * An instance of a DCS scenery object
  */
 message Scenery {
+  // The DCS type-name of the scenery
+  string type = 1;
+  // The name of the scenery
+  string name = 2;
+  // The position of the scenery
+  Position position = 3;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


justin-lovell (approved):

protos/dcs/common/v0/common.proto
```diff
@@ -264,12 +264,28 @@ message Weapon {
...
+  // The DCS generated ID
+  uint32 id = 1;
+  // The DCS type-name of the static
+  string type = 2;
+  // The name of the static
+  string name = 3;
+  // The coalition the static belongs to
+  Coalition coalition = 4;
+  // The position of the static
+  Position position = 5;
```
justin-lovell, line 276: A minor note - and we can evolve this in a later PR - cargo would have additional properties such as weight.

If statics are being transported, or is attached to a moving object (e.g. statics on CV), statics would have a velocity.

PR #108, opened 2022-01-15
WIP - Tracking of equivalent functions

justin-lovell:
Initial commit to track equivalence and scope of DCS API's

At some point, I would like to refer to the completeness of the gRPC API surface area. This would allow the scope to be worked on incrementally.

Feel free to suggest updates to the paragraph of not committing to a roadmap. There are a lot of functions which should not be implemented.

PR #124, opened 2022-02-07
Follow DCS Hierarchy for Object Inheritance

justin-lovell:
All the object types within DCS follow an inheritance chain, therefore
creating a class hierarchy.

We are able to use a builder method to allow the attributes of the chain
to ensure that no common properties are overlooked.

This was prompted when doing the inventory taking upon the following PR: https://github.com/DCS-gRPC/rust-server/pull/108

### DIFF
--- a/lua/DCS-gRPC/exporters/object.lua
+++ b/lua/DCS-gRPC/exporters/object.lua
@@ -5,6 +5,13 @@
 -- be kept in sync
 --
 
+local function mergeTable(t1, t2)
+  for k,v in pairs(t2) do
+    t1[k] = v
+  end
+  return t1
+end
+
 -- Convert DCS's unusual right-hand coordinate system where +x points north to a more common
 -- left-hand coordinate system where +z points north (and +x points east).
 GRPC.exporters.vector = function(v)
@@ -24,8 +31,8 @@ GRPC.exporters.position = function(pos)
   }
 end
 
-GRPC.exporters.unit = function(unit)
-  local vector = unit:getVelocity()
+local function exportObject(object, extendedObject)
+  local vector = object:getVelocity()
 
   local heading = math.deg(math.atan2(vector.z, vector.x))
   if heading < 0 then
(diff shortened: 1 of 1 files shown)


justin-lovell:
Based on the documentation here: https://wiki.hoggitworld.com/view/DCS_Class_Object, this has got me thinking if we need to relook at how the Protobuf message is composed. Currently, it is very flat.

What if we kicked up to the following as a composition bridge? (ignore field identifers; only retaining it for the refactor)

```diff
+message ObjectExtension {
+  string type = 2;
+  string name = 3;
+  Position position = 5;
+}
+message CoalitionExtension {
+   Coalition coalition = 4;
+}
message Static {
  uint32 id = 1;
-  string type = 2;
-  string name = 3;
-  Coalition coalition = 4;
-  Position position = 5;
+ ObjectExtension object = 99;
+ CoalitionExtension coalition = 100;
}
```

Alternatively, I could easily build up the common attributes on the five kinds of objects DCS contains.

I'm kind of leaning towards the latter, but wanted to check if the alternative was preferred?

rurounijones:
What is the reasoning for doing this? Adding a hierarchy here adds no value I think at the protobuf level/public API level and I don't really see any advantages in the lua level either. It reduces lines of code but I don't think it simplifies anything.

justin-lovell:
I've noticed a few missing fields across the different object types (e.g. `speed` and `heading` not transposed onto `Static`). I'll give some thought on how to scaffold it and KISS.

rurounijones:
I didn't add `speed` or `heading` to statics because, well... they are static. there is no value to adding them. I guess you could argue that `heading` may have a use but I cannot think of one off the top of my head.

justin-lovell:
The use case would be statics/cargo (they are the "same") attached to ships and/or hooked to a helicopter :D

justin-lovell:
I'm closing the PR for favor of moving the branch to the main repository.

In consideration, I've decided I am going to knock out the various base attributes onto the various object types -- it would be quick to get the base stuff out, and then we can fill out the attributes on the sealed classes incrementally.

PR #125, opened 2022-02-08
Tracking of equivalent functions

justin-lovell:
Originally started, but changed repository: https://github.com/DCS-gRPC/rust-server/pull/108

Initial commit to track equivalence and scope of DCS API's

At some point, I would like to refer to the completeness of the gRPC API surface area. This would allow the scope to be worked on incrementally.

Feel free to suggest updates to the paragraph of not committing to a roadmap. There are a lot of functions which should not be implemented.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -1,0 +1,556 @@
+# DCS Functional Parity
+
+The DCS gRPC Server API implementation is still a work in progress, and therefore the following API functions have
+been implemented.
+
+The unimplemented functions is not committed to a roadmap. We do envision full DCS API equivalence at some point in the
+future; pull requests are welcomed for expanding the API equivalency.
+
+---
+
+## Singletons
+
+### Atmosphere Singleton
+- [x] `getWind`
+- [x] `getWindWithTurbulence`
+- [x] `getTemperatureAndPressure`
+
+### Coalitions Singleton
+- [ ] `addGroup`
+  - [ ] Sea
+    - [ ] Group
+    - [ ] Units
+    - [ ] Waypoints
+    - [ ] Tasks
+  - [ ] Ground
+    - [ ] Group
+    - [x] Units
+    - [ ] Waypoints
+    - [ ] Tasks
+  - [ ] Plane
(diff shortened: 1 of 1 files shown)


STATUS.md
```diff
@@ -0,0 +1,489 @@
...
+#### GameGUI API
+  - [ ] `load_mission`
+  - [ ] `load_next_mission`
+
+
+### Timer Service
+- [x] `getTime`
+- [x] `getAbsTime`
+- [x] `getTime0`
+- [ ] `scheduleFunction` (TODO -- should this be implemented?)
```
justin-lovell, line 121: Transposing @rurounijones remark from initial draft PR.

> I would be against it. Clients can all the APIs when they need to,

justin-lovell: If the server is populated, I agree that this seems irrelevant.

However, in the use cases of servers with sparse density, the `scheduleFunction` could be a useful trigger to state "how much model time has really lapsed" when servers pause/resume as players connect/disconnect.

I would prefer to find an alternative solution to the pause/unpause scenario and tracking model/mission time.

rurounijones: Since it is marked as a TODO we can deal with this later.

STATUS.md
```diff
@@ -0,0 +1,489 @@
...
+- [Options](https://wiki.hoggitworld.com/view/DCS_func_setOption)
+  - [ ] `ROE`
+  - [ ] `Reaction To Threat`
+  - [ ] `Radar Using`
+  - [ ] `Flare Using`
+  - [ ] `Formation`
+  - [ ] `RTB On Bingo`
+  - [ ] `silence`
+  - [ ] `Disperse on Attack`
+  - [ ] `Alarm State`
```
rurounijones, line 459: This has been implemented: https://github.com/DCS-gRPC/rust-server/blob/status/protos/dcs/controller/v0/controller.proto#L8

justin-lovell:
As up to date as I can get it up to - let's merge!

rurounijones (approved):

PR #129, opened 2022-04-03
Add more detail to `GroupCommand` event.

rurounijones:
Replace `groupName` field in the `GroupCommand` event with all the group
details as exposed by the group exporter (currently id, name, coalition
and category).

This change was made based on experience writing a client that processes
these events where only having the groupName was a limitation.

This change breaks backwards compatibility with version 0.4.0 where the
`GroupCommand` event was first added.

### DIFF
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -437,8 +437,8 @@ message StreamEventsResponse {
   }
 
   message GroupCommandEvent {
-    // The name of the group of the player who ran the command
-    string group_name = 1;
+    // Details of the group to which the player who ran the command is a unit of
+    dcs.common.v0.Group group = 1;
     // A struct containing details of the command that was run by a player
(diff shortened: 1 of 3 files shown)


protos/dcs/mission/v0/mission.proto
```diff
@@ -660,6 +660,7 @@ message RemoveCoalitionCommandItemResponse {
 // be emitted to all connected DCS-gRPC clients for processing as they see fit.
 // The emitted event will include the group name.
 message AddGroupCommandRequest {
+
```
rkusa, line 663: Change made on purpose?

rkusa (dismissed):

protos/dcs/mission/v0/mission.proto
rurounijones: Whoops. Fixed

rkusa (approved):

PR #133, opened 2022-04-15
Add more detail to `GroupCommand` event.

rurounijones:
Replace `groupName` field in the `GroupCommand` event with all the group
details as exposed by the group exporter (currently id, name, coalition
and category).

This change was made based on experience writing a client that processes
these events where only having the groupName was a limitation.

This change breaks backwards compatibility with version 0.4.0 where the
`GroupCommand` event was first added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


rkusa:
I am confused, isn't this the same as https://github.com/DCS-gRPC/rust-server/pull/129? :D

rurounijones:
Oh goddamnit... yes, yes it is. My bad

PR #134, opened 2022-04-15
Add various Command and Control APIs

rurounijones:
Add various APIs that are suited to server command and control. These also implement some of the APIs required for LotATC support.

### DIFF
--- a/lua/DCS-gRPC/methods/hook.lua
+++ b/lua/DCS-gRPC/methods/hook.lua
@@ -57,3 +57,38 @@ end
 GRPC.methods.isServer = function()
   return GRPC.success({server = DCS.isServer()})
 end
+
+GRPC.methods.banPlayer = function(params)
+  if params.id == 1 then
+    return GRPC.errorInvalidArgument("Cannot ban the server user")
+  end
+
+  local player_id = net.get_player_info(params.id, "id")
+
+  if not player_id then
+    return GRPC.errorNotFound("Could not find player with the ID of " .. params.id)
+  end
+
+  return GRPC.success({banned = net.banlist_add(params.id, params.period, params.reason)})
+end
+
+GRPC.methods.unbanPlayer = function(params)
+  return GRPC.success({unbanned = net.banlist_remove(params.ucid)})
+end
+
+GRPC.methods.getBannedPlayers = function()
+  local result = {}
+
+  for i, detail in ipairs(net.banlist_get()) do
+    result[i] = {
+      ucid = detail.ucid,
+      ipAddress = detail.ipaddr,
+      playerName = detail.name,
+      reason = detail.reason,
+      bannedFrom = detail.banned_from,
+      bannedUntil = detail.banned_until
+    }
+  end
+
+  return GRPC.success({bans = result})
+end
--- a/lua/DCS-gRPC/methods/hook.lua
+++ b/lua/DCS-gRPC/methods/hook.lua
@@ -27,6 +27,11 @@ GRPC.methods.stopMission = function()
   return GRPC.success(nil)
 end
 
+GRPC.methods.exitProcess = function()
+  DCS.exitProcess()
+  return GRPC.success(nil)
+end
+
 GRPC.methods.hookEval = function(params)
   local fn, err = loadstring(params.lua)
   if not fn then
--- a/lua/DCS-gRPC/methods/hook.lua
+++ b/lua/DCS-gRPC/methods/hook.lua
@@ -13,6 +13,20 @@ GRPC.methods.getMissionFilename = function()
   return GRPC.success({name = DCS.getMissionFilename()})
 end
 
+GRPC.methods.getPaused = function()
+  return GRPC.success({paused = DCS.getPause()})
+end
+
+GRPC.methods.setPaused = function(params)
+  DCS.setPause(params.paused)
+  return GRPC.success(nil)
+end
+
+GRPC.methods.stopMission = function()
+  DCS.stopMission()
+  return GRPC.success(nil)
+end
+
 GRPC.methods.hookEval = function(params)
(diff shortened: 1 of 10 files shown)


src/rpc/hook.rs
```diff
@@ -27,4 +59,12 @@ impl HookService for HookRpc {
         })?;
         Ok(Response::new(hook::v0::EvalResponse { json }))
     }
+
+    async fn exit_process(
+        &self,
+        request: Request<hook::v0::ExitProcessRequest>,
+    ) -> Result<Response<hook::v0::ExitProcessResponse>, Status> {
+        self.notification("exitProcess", request).await?;
+        Ok(Response::new(hook::v0::ExitProcessResponse {}))
```
rkusa, line 68: In case this is not WIP, there is no Lua counterpart for `exitProcess` in the PR yet

rurounijones: No need to review these yet. I haven't even tested that they work (which would have uncovered this 😁 )

Testing it a bit of a PITA since you have to fire up DCS etc so I prefer to write a bunch of these simple APIs then test them all together, fix any issues, push then move the PR out of draft mode.

lua/DCS-gRPC/methods/net.lua
```diff
@@ -52,3 +52,18 @@ GRPC.methods.forcePlayerSlot = function(params)
...
+  end
+
+  local player_id = net.get_player_info(params.id, "id")
+
+  if not player_id then
+    return GRPC.errorNotFound("Could not find player with the ID of " .. params.id)
+  end
+
+  net.kick(params.id, params.message)
+  return GRPC.success(nil)
```
rurounijones, line 68: I cannot actually get this API to return an error response for reasons I am not sure about (I am sure I am missing something) however it _does_ appear to exit the method, it just doesn't return an error response.

This is "good enough" at the moment and I don't want to hold up a release for it.

justin-lovell: The reason is that the Rust code does not wait on the response from the Lua function run.

Here is an example for you to reference. I did spot a few other API's which does not consider the error response as well, so we will have to do a sweep for them at some time in the future.

https://github.com/DCS-gRPC/rust-server/blob/2db3b3bf4b1be0f7628a8171356198770d59dd89/src/rpc/atmosphere.rs#L8-L18

justin-lovell (approved):

lua/DCS-gRPC/methods/net.lua
rkusa: The `?` at the end of line 12 in the snippet above should be all that is needed to handle the error. I cannot spot anything wrong with the code. I'll test it once I am back on my PC (later tomorrow probably) to look into it.

rurounijones: In my code I use the notification method because there is no response and I forgot that doing so would also swallow the error message. So there isn't anything for you to do here I think. I just need to fix the code I am using

rurounijones:
Updated the KickPlayer rust code so that it returns a result (or error, in this case)

lua/DCS-gRPC/methods/net.lua
rkusa: Ah gotcha, I misunderstood @justin-lovell example - it was an example where the error is handled. We can also remove the `self.notification()` method and use `self.request()` everywhere. Might be a good idea for a consistent behavior?

lua/DCS-gRPC/methods/hook.lua
```diff
@@ -22,3 +49,46 @@ GRPC.methods.hookEval = function(params)
...
+
+  for i, detail in ipairs(net.banlist_get()) do
+    result[i] = {
+    ucid = detail.ucid,
+    ipAddress = detail.ipaddr,
+    playerName = detail.name,
+    reason = detail.reason,
+    bannedFrom = detail.banned_from,
+    bannedUntil = detail.banned_until
+    }
```
rkusa, line 90: Some indentation consistency: 
```suggestion
    result[i] = {
      ucid = detail.ucid,
      ipAddress = detail.ipaddr,
      playerName = detail.name,
      reason = detail.reason,
      bannedFrom = detail.banned_from,
      bannedUntil = detail.banned_until
    }
```

protos/dcs/hook/v0/hook.proto
```diff
@@ -8,9 +8,45 @@ service HookService {
...
   rpc Eval(EvalRequest) returns (EvalResponse) {}
+
+  // https://wiki.hoggitworld.com/view/DCS_func_exitProcess
+  rpc ExitProcess(ExitProcessRequest) returns (ExitProcessResponse) {}
+
+  // https://wiki.hoggitworld.com/view/DCS_func_isMultiplayer
+  rpc IsMultiplayer(IsMultiplayerRequest) returns (IsMultiplayerResponse) {}
+
+  // https://wiki.hoggitworld.com/view/DCS_func_isServer
+  rpc IsServer(IsServerRequest) returns (IsServerResponse) {}
```
rkusa, line 39: Wouldn't this always be `true` anyway, as the server is only ever started for servers (either being SP or MP)?

rurounijones: Probably, but since this is a pass-through to the DCS API I am leaving the nuances of the implementation to them.

lua/DCS-gRPC/methods/net.lua
justin-lovell: I agree @rkusa -- if we do decide to have a async processor message box style in the future, and then the outcome of the notification should be read. At this moment, that API style feels too much overhead when most of the API's should be immediate request/response for simplicity.

justin-lovell (approved):

rkusa (approved):

PR #136, opened 2022-04-19
remove ipc notification usages

rkusa:
IPS `request`s will be used instead. This ensures that errors are
handled everywhere (they where ignored for notifications).

Result of the discussion in https://github.com/DCS-gRPC/rust-server/pull/134#discussion_r852141047

### DIFF
--- a/lua/DCS-gRPC/methods/controllers.lua
+++ b/lua/DCS-gRPC/methods/controllers.lua
@@ -28,5 +28,5 @@ GRPC.methods.setAlarmState = function(params)
 
   controller:setOption(state_id, params.alarmState - 1)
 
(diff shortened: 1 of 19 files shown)


rurounijones:
If the aim is to standardise then shouldn't we standardise on all rust code returning the value from the lua code? It means we have to have `GRPC.success(nil)` in all the lua methods but I think we have that anyway.

Right now we still have the notification being removed but there are still two ways of structuring the rust code.

rkusa:
@rurounijones Good point! I've updated all methods to be consistent. This also required to replace all `GRPC.success(nil)` with `GRPC.success({})`.

rurounijones (approved):

PR #140, opened 2022-05-04
Add `MarkupToAll` and `MarkupToCoalition` APIs

rurounijones:
Although there is no `MarkupToCoalition` in the scripting environment;
we will add a gRPC level API for one to keep in line with the `MarkTo*`
APIs. There is no way to do a `MarkupToGroup` so we are not implementing
it.


justin-lovell (approved):

justin-lovell:
I like the API footprint of breaking out the "all" and "coalition" ... an improvement over ED's naming.

PR #144, opened 2022-05-08
Add Static Objects API

justin-lovell:
Implementation of `addStaticObject` via the gRPC interface.

The cargo type is free text so to avoid the requirements of DCS-gRPC
installations and clients to be coupled to any DCS updates / new units /
 renames

* Static positioning
* Linked unit
* Cargo handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `OutTextForUnit` API
+- `AddStaticObject` API (for standard static objects)
+- `AddLinkedStatic` API (for statics linked to units such as ships)
 - `MarkupToAll` API
 - `MarkupToCoalition` API
 - `GetTheatre` API
--- a/lua/DCS-gRPC/methods/coalitions.lua
+++ b/lua/DCS-gRPC/methods/coalitions.lua
@@ -120,6 +120,91 @@ GRPC.methods.addGroup = function(params)
   return GRPC.success({group = GRPC.exporters.group(Group.getByName(template.name))})
 end
 
+GRPC.methods.addStaticObject = function(params)
+  if params.name == nil or params.name == "" then
+    return GRPC.errorInvalidArgument("name not supplied")
+  end
+  if params.type == nil or params.type == "" then
+    return GRPC.errorInvalidArgument("type not supplied")
+  end
+  if params.country_id == 0 or params.country_id == 15 then
+    return GRPC.errorInvalidArgument("invalid country code")
+  end
+
+  if params.position == nil then
+    return GRPC.errorInvalidArgument("provide position")
+  end
+
+  local pos = coord.LLtoLO(params.position.lat, params.position.lon)
+  local staticTemplate = {
+    name = params.name,
+    type = params.type,
+    heading = math.rad(params.heading),
+    dead = params.dead,
+    x = pos.x,
+    y = pos.z,
+  }
+
+  if params.score ~= nil then
+    staticTemplate.rate = params.score
+  end
+
+  if params.cargoMass > 0 then
+    staticTemplate.canCargo = true
+    staticTemplate.mass = params.cargoMass
+  end
+
+  coalition.addStaticObject(params.country - 1, staticTemplate)
+
+  return GRPC.success({ name = params.name })
+end
+
+GRPC.methods.addLinkedStatic = function(params)
+  if params.name == nil or params.name == "" then
+    return GRPC.errorInvalidArgument("name not supplied")
+  end
+  if params.type == nil or params.type == "" then
+    return GRPC.errorInvalidArgument("type not supplied")
+  end
+  if params.country_id == 0 or params.country_id == 15 then
+    return GRPC.errorInvalidArgument("invalid country code")
+  end
+  if params.unit == nil or params.unit == "" then
+    return GRPC.errorInvalidArgument("provide the unit to ")
+  end
+
+  local unit = Unit.getByName(params.unit)
+  if unit == nil then
+    return GRPC.errorNotFound("offset unit name not found")
+  end
+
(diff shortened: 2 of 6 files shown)


protos/dcs/coalition/v0/coalition.proto
```diff
@@ -121,6 +124,39 @@ message AddGroupResponse {
...
+
+  message Offset {
+    // the name of the unit to offset from
+    string unit = 1;
+    double x = 2;
+    double y = 3;
+  }
+
+  // cargo mass in kilograms
+  uint32 cargo_mass = 10;
```
justin-lovell, line 153: There are a lot variations on the behavior, depending on the scenario. Examples are:

* Heading/angle is orientated based off the linked unit's heading
* Cargo is only allowed on non-linked statics

An alternative I was debating was the following pseudo:
```
oneof position {
   StaticPositioning static = YYY;
   LinkedPositioning offset = YYYY;
}

message StaticPositioning {
  double heading = 7;
  dcs.common.v0.Position position = 8;
  uint32 cargo_mass = 10;
}

message Offset {
  // the name of the unit to offset from
  string unit = 1;
  double x = 2;
  double y = 3;
}
```
Alternatively, two discrete API's for `AddStatic` and `AddLinkedStatic`

rurounijones: From what I understand by looking at the hoggitworld wiki it looks like the offset / linking stuff is to put statics on ships... carriers, really with planes, cranes etc.

Given this fact and my impression that this "ship linking" offset stuff will be used a _lot_ less than non-linking stuff I think your suggestion of two separate APIs is a good one.

lua/DCS-gRPC/methods/coalitions.lua
```diff
@@ -120,6 +120,70 @@ GRPC.methods.addGroup = function(params)
...
+  else
+    local pos = coord.LLtoLO(params.position.lat, params.position.lon)
+    staticTemplate.x = pos.x
+    staticTemplate.y = pos.z
+  end
+
+  coalition.addStaticObject(params.country - 1, staticTemplate)
+
+  return GRPC.success({
+    static = GRPC.exporters.static(StaticObject.getByName(params.name)),
```
rurounijones, line 183: This may cause race condition issues. I have seen in the past that immediate retrieve calls in the same simulation frame can fail because the game hasn't fully instantiated the object yet.

It might be safer to not return anything. The `DCS-gRPC` client knows the details since it passed it all in apart from the ID and we tend not to use the object IDs in our APIs but use the names instead, which is known.

justin-lovell: You've got more experience with high-load servers, so I will lean on your experience and adjust. We may also want to adjust the following too then: https://github.com/DCS-gRPC/rust-server/blob/aa68abcb1d18cb0cc8c087362dbea9f5a2141c4b/lua/DCS-gRPC/methods/coalitions.lua#L120

PS: I agree with adding the fetch statics. It's next on my todo, I want to reduce the number of concurrent PR's.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `OutTextForUnit` API
+- `AddStaticObject` API  (for standard static objects)
```
rurounijones, line 11: Extra space before the bracket.

rurounijones (approved):
Overall I think this is fine to ship if we want. I have some small comments that are more suggestions based on taste than anything else.

protos/dcs/coalition/v0/coalition.proto
```diff
@@ -9,6 +9,16 @@ service CoalitionService {
...
+  // Focussed on statics (linked statics - see `AddLinkedStatic`)
+  // https://wiki.hoggitworld.com/view/DCS_func_addStaticObject
+  rpc AddStaticObject(AddStaticObjectRequest)
+      returns (AddStaticObjectResponse) {}
+
+  // Focussed on properties relevant to linked static objects
+  // https://wiki.hoggitworld.com/view/DCS_func_addStaticObject
+  rpc AddLinkedStatic(AddLinkedStaticRequest)
+      returns (AddLinkedStaticResponse) {}
+
```
rkusa, line 21: I'd personally prefer having only one `AddStaticObject` method and distinguish between linked and not linked with a `oneof` inside of the payload. But I could live with having the methods, too.

rkusa (commented):
👍

protos/dcs/coalition/v0/coalition.proto
```diff
@@ -121,6 +131,58 @@ message AddGroupResponse {
...
+  // string name of the livery for the aircraft
+  string livery = 4;
+  // boolean for whether or not the object will appear as a wreck
+  bool dead = 5;
+  // number value for the "score" of the object when it is killed
+  optional uint32 score_rate = 6;
+  // the name of the unit to offset from
+  string unit = 7;
+  // the angle to relative to the linked unit
+  double angle = 8;
```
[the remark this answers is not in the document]
rkusa: Or if we consider renaming it, how about: `orientation`?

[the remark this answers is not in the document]
justin-lovell: Other ideas/suggestions:

`direction`
`alignment_angle`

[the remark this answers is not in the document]
rurounijones: I guess the name is less important than we make sure the documentation on the field explicitly calls out what the values mean because I am still not 100% sure. For example I am assuming clockwise heading offset but can we do negatives? is -90 and 270 the same thing?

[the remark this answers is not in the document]
justin-lovell: I tested with negative values. It is anti-clockwise.

rurounijones (approved):

rkusa (approved):

PR #145, opened 2022-05-08
Prepend Search Path

justin-lovell:
Reduces the number of search entries for search paths.

Not all modules would append a semi-colon (`;`) for the path at the end.
 Inserting the gRPC search path in the beginning would eliminate a
 potential blank search path `somePath;;pathForGRPC;` and instead make
 the brief search path of `pathForGRPC;someOtherPath`.

### DIFF
--- a/lua/DCS-gRPC/grpc-hook.lua
+++ b/lua/DCS-gRPC/grpc-hook.lua
@@ -4,7 +4,7 @@ local function load()
 
   -- Let DCS know where to find the DLLs
   if not string.find(package.cpath, GRPC.dllPath) then
-    package.cpath = package.cpath .. [[;]] .. GRPC.dllPath .. [[?.dll;]]
+    package.cpath =  GRPC.dllPath .. [[?.dll;]] .. package.cpath
   end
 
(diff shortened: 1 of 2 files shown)


justin-lovell:
Optional alternative implementation -- no biggie if not adopted.

rurounijones (changes requested):
I am not completely against this but I tend to avoid prepending since, in my experience, it can violate the principle of least surprise that people have in their assumptions that the order in which they include / load code / modules (in any language) is maintained in how that code / module adds itself to shared paths / resources.

I'll let @rkusa cast the deciding vote.

rkusa:
I am in favour of keeping it as is and append our path. If we want to avoid the empty search path, we could do something like:

```lua
if package.cpath ~= "" and string.sub(package.cpath, -1) ~= "~" then
  package.cpath = package.cpath .. ";"
end
package.cpath = package.cpath .. GRPC.dllPath .. [[?.dll;]]
```

Though I'd personally probably just live with the empty search path, but I am also not opposed to the additional logic to avoid it.

PR #146, opened 2022-05-10
Upgrade dependencies and bundle protoc.exe

rkusa:
Upgrade dependencies. This includes major upgrades. Unfortunately, the dependency that generates code out of proto files isn't bundling `protoc` anymore, so I am afraid we have to bundle it ourselves (or keep using the older version). Anyone against bundling `protoc.exe` in the repo?


src/services.rs
```diff
@@ -1,118 +0,0 @@
...
-        } else if path.starts_with(WorldServiceServer::<MissionRpc>::NAME) {
-            self.world.call(req)
-        } else {
-            Box::pin(std::future::ready(Ok(http::Response::builder()
-                .status(200)
-                .header("grpc-status", "12")
-                .header("content-type", "application/grpc")
-                .body(tonic::codegen::empty_body())
-                .unwrap())))
-        }
```
rkusa, line 116: With the dependency upgrades, this code wasn't working anymore. It was always returning the `else` path. The code was initially added due to compile time issues (in https://github.com/DCS-gRPC/rust-server/pull/61). Fortunately, the compile time issues are gone (either due to a Rust or dep update). At least for me - if you have the time, please check the compile time for you too. This is why I've removed the workaround again instead of fixing it.

rkusa:
Converted back to draft as I have to figure out a way to bundle `protoc.exe` for both Windows and Linux (CI) ... any ideas? @rurounijones @justin-lovell

justin-lovell:
For CI, perhaps use - 
```
apt install -y protobuf-compiler
```

I would anticipate that for the local development experience, have a section for requiring the tool to be installed - I would encourage with Chocolately https://community.chocolatey.org/packages/protoc

rurounijones:
Agree with the above suggesions by @justin-lovell

rkusa:
The `protoc` version for ubuntu is unfortunately too outdated, so I've ended up downloading a binary for linux from Github. Sorry for the CI try-and-error commit spam.

rkusa:
> I would anticipate that for the local development experience, have a section for requiring the tool to be installed - I would encourage with Chocolately https://community.chocolatey.org/packages/protoc

I am personally not a huge fan of any additional step to simply running `cargo build`. It additionally has the drawback that we don't control the version. So if we don't want to bundle the `protoc.exe` in the repo, the only alternative I can think of is to automatically download it as part of the `cargo build`.

rurounijones:
> I am personally not a huge fan of any additional step to simply running `cargo build`. It additionally has the drawback that we don't control the version. So if we don't want to bundle the `protoc.exe` in the repo, the only alternative I can think of is to automatically download it as part of the `cargo build`.

In rubyland there is a convention that you can run a "setup" command that does the first-time installation of stuff like this. Can we do the same here? Have a `cargo setup` that sets up the project dependencies so the full first-time steps are

```
cargo install
cargo setup
cargo build
```

Otherwise adding it to the `cargo build` stage where it checks for `protoc.exe` and downloads it if it is missing would be my preferred choice.

justin-lovell:
> the only alternative I can think of is to automatically download it as part of the cargo build.

I like that alternative. Is your thinking similar to [this build.rs](https://doc.rust-lang.org/cargo/reference/build-script-examples.html)? It looks neat.

rkusa:
Thanks for the feedback!

@rurounijones Such a `setup` convention does not exist in the Rust community (AFAIK).
@justin-lovell Yes, the download would be done in a `build.rs`. I'd add it to https://github.com/DCS-gRPC/rust-server/blob/main/stubs/build.rs.

I looked into downloading it, but after adding 5 dependencies just to download and extract protoc, I didn't like the idea too much anymore.

> Alternatively (and I know this will probably be too much work): Create our own rust Crate that does include the dependency and depend on that crate.

Which made this look like a way better/easier solution. I've updated the PR accordingly. I've created the dependency under my own Github user for now https://github.com/rkusa/protoc-bundled. We could also move it to `DCS-gRPC` if we like.

I'd actually tend to keeping it as a Git dependency and not publish it to https://crates.io/. Instead of rolling our own dependency, we could also use something like https://lib.rs/crates/protoc-bin-vendored, but I am not too keen on pulling executables from yet another stranger.

In the future we could move to building protoc instead of bundling/downloading it. This currently requires `cmake` to be available, which brings us back to the same issue. However, a removal of the `cmake` dependencies for building seems to be in the work (https://github.com/tokio-rs/prost/pull/620).

What do you think?

rurounijones (approved):
The current dependency in a separate dependency only accessible via git sounds like a good solution.

justin-lovell (dismissed):
learned a lot from this PR. cheers.

src/services.rs
rurounijones: I will test this tomorrow locally

rkusa: @rurounijones I'll wait for your feedback before merging 👍

rurounijones: Am receiving the following errors on this branch and this branch only. I re-ran on main branch without issues.

I chopped off only to the last error and message. Lemme know if you need the entire output


```
   C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(245,5): error MSB8066: Custom build for 'C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\CMakeFiles\3800ac9a10098988c9e6c56711b08d23\map_lite_unittest.pb.cc.rule;C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\CMakeFiles\3800ac9a10098988c9e6c56711b08d23\unittest_import_lite.pb.cc.rule;C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\CMakeFiles\3800ac9a10098988c9e6c56711b08d23\unittest_import_public_lite.pb.cc.rule;C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\CMakeFiles\3800ac9a10098988c9e6c56711b08d23\unittest_lite.pb.cc.rule;C:\Users\jeff\.cargo\registry\src\github.com-1ecc6299db9ec823\prost-build-0.10.3\third-party\protobuf\cmake\CMakeLists.txt' exited with code 3. [C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\lite-test.vcxproj]
    Generating Code...
    test_plugin.vcxproj -> C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-68bc37c7b395e905\out\build\Debug\test_plugin.exe

  --- stderr
  CMake Warning:
    Manually-specified variables were not used by the project:

      CMAKE_ASM_FLAGS
      CMAKE_ASM_FLAGS_DEBUG


  thread 'main' panicked at '
  command did not execute successfully, got: exit code: 1

  build script failed, must exit now', C:\Users\jeff\.cargo\registry\src\github.com-1ecc6299db9ec823\cmake-0.1.48\src\lib.rs:975:5
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

rkusa: Hmm, it looks like it tries to build protoc for you instead of using the bundled one. If I am reading [the code](https://github.com/tokio-rs/prost/blob/master/prost-build/build.rs#L96) correctly, it should prefer the bundled one over compiling it from source. The difference between our systems might be that I don't have `cmake` installed.

From the error I am assuming you are using Visual Studio for the build, could you try if you get the same error when simply running `cargo build` from any console? This might help to narrow the cause of the error down.

rurounijones: From a powershell console. I have no idea where `"-G" "Visual Studio 17 2022` is coming from.

```
cargo build
   Compiling prost-build v0.10.3
error: failed to run custom build command for `prost-build v0.10.3`

Caused by:
  process didn't exit successfully: `C:\Users\jeff\source\repos\DCS-gRPC\rust-server\target\debug\build\prost-build-7115271569e6896d\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=C:\Users\jeff\.cargo\registry\src\github.com-1ecc6299db9ec823\prost-build-0.10.3\third-party\protobuf\cmake
  CMAKE_TOOLCHAIN_FILE_x86_64-pc-windows-msvc = None
  CMAKE_TOOLCHAIN_FILE_x86_64_pc_windows_msvc = None
  HOST_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_x86_64-pc-windows-msvc = None
  CMAKE_GENERATOR_x86_64_pc_windows_msvc = None
  HOST_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  CMAKE_PREFIX_PATH_x86_64-pc-windows-msvc = None
  CMAKE_PREFIX_PATH_x86_64_pc_windows_msvc = None
  HOST_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_x86_64-pc-windows-msvc = None
  CMAKE_x86_64_pc_windows_msvc = None
  HOST_CMAKE = None
  CMAKE = None
  running: "cmake" "C:\\Users\\jeff\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\prost-build-0.10.3\\third-party\\protobuf\\cmake" "-G" "Visual Studio 17 2022" "-Thost=x64" "-Ax64" "-DCMAKE_INSTALL_PREFIX=C:\\Users\\jeff\\source\\repos\\DCS-gRPC\\rust-server\\target\\debug\\build\\prost-build-68bc37c7b395e905\\out" "-DCMAKE_C_FLAGS= -nologo -MD -Brepro" "-DCMAKE_C_FLAGS_DEBUG= -nologo -MD -Brepro" "-DCMAKE_CXX_FLAGS= -nologo -MD -Brepro" "-DCMAKE_CXX_FLAGS_DEBUG= -nologo -MD -Brepro" "-DCMAKE_ASM_FLAGS= -nologo -MD -Brepro" "-DCMAKE_ASM_FLAGS_DEBUG= -nologo -MD -Brepro" "-DCMAKE_BUILD_TYPE=Debug"

  --- stderr
  thread 'main' panicked at '
  failed to execute command: The system cannot find the file specified. (os error 2)
  is `cmake` not installed?

  build script failed, must exit now', C:\Users\jeff\.cargo\registry\src\github.com-1ecc6299db9ec823\cmake-0.1.48\src\lib.rs:975:5
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

rkusa: After running `cargo clean` and then trying to build again, I am encountering the same error. I am a bit confused that this wasn't caught by CI. Anyway, I've pushed a ~fix~ workaround. Should work now. I btw. also have the `"-G" "Visual Studio 17 2022` in there, so probably normal on Windows.

Would you mind giving it another try, too?

rurounijones: Confirmed all good

rurounijones (approved):

Issue #137, opened 2022-04-20
Eval failure


rurounijones:
Noting here for reference.

Running the following via the gRPC hook eval API `terrain.getRadio()[1]` returns a gRPC error response with the following in the `dcs.log`.

```
2022-04-20 11:45:00.730 ERROR   [GRPC-Hook]: Error retrieving next command: callback error
stack traceback:
	[C]: in function 'next'
	[string "C:\Users\jeff\source\repos\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua"]:244: in function <[string "C:\Users\jeff\source\repos\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua"]:242>
	[C]: in function 'pcall'
	[string "C:\Users\jeff\source\repos\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua"]:259: in function 'onSimulationFrame'
	[string "C:\Users\jeff\source\repos\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc-hook.lua"]:34: in function <[string "C:\Users\jeff\source\repos\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc-hook.lua"]:32>
	[C]: in function 'pcall'
	[string "MissionEditor/loadUserScripts.lua"]:131: in function <[string "MissionEditor/loadUserScripts.lua"]:126>
	(tail call): in function <(tail call):-1>
caused by: Failed to deserialize result for method hookEval: deserialize error: invalid type: integer `2`, expected a string key
{
  "role" = {
    1 = "ground",
    2 = "tower",
    3 = "approach",
  },
  "callsign" = {
    1 = {
      "common" = {
        1 = "OMAA",
        2 = "OMAA",
      },
    },
  },
  "sceneObjects" = {
    1 = "t:60392560",
  },
  "radioId" = "airfield22_0",
  "frequency" = {
    2 = {
      1 = 0,
      2 = 119200000,
    },
  },
}
```

rurounijones:
Grimes pointed out that the actual lua table contains both string and numeric integer keys which  I think is causing the failure above:

```
[...]

PR #147, opened 2022-05-10
Fix eval json encoding limitations

rkusa:
Here is a possible fix for https://github.com/DCS-gRPC/rust-server/issues/137. It's fixed by using `net.lua2json` instead of doing the Lua->JSON conversion in Rust.

As an alternative, I might also be able to update the Rust Lua->JSON conversion to handle the cases mentioned in #137, but simply using `lua2json` is a way easier solution.


rurounijones (approved):
Sweeet

PR #148, opened 2022-05-15
Get Static Objects API

justin-lovell:
Retrieves all the statics populated to a coalition (or all!)

Tracks positions of statics attached to units and/or cargo as expected.

### DIFF
--- a/lua/DCS-gRPC/methods/coalitions.lua
+++ b/lua/DCS-gRPC/methods/coalitions.lua
@@ -120,6 +120,21 @@ GRPC.methods.addGroup = function(params)
   return GRPC.success({group = GRPC.exporters.group(Group.getByName(template.name))})
 end
 
+GRPC.methods.getStaticObjects = function(params)
+  local result = {}
+  for _, coalitionId in pairs(coalition.side) do
+    if params.coalition == 0 or params.coalition - 1 == coalitionId then -- Decrement for non zero-indexed gRPC enum
+      local staticObjects = coalition.getStaticObjects(coalitionId)
+
+      for _, staticObject in ipairs(staticObjects) do
+        table.insert(result, GRPC.exporters.static(staticObject))
+      end
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


lua/DCS-gRPC/methods/coalitions.lua
```diff
@@ -120,6 +120,21 @@ GRPC.methods.addGroup = function(params)
...
+GRPC.methods.getStaticObjects = function(params)
+  local result = {}
+  for _, coalitionId in pairs(coalition.side) do
+    if params.coalition == 0 or params.coalition - 1 == coalitionId then -- Decrement for non zero-indexed gRPC enum
+      local staticObjects = coalition.getStaticObjects(coalitionId)
+
+      for _, staticObject in ipairs(staticObjects) do
+        table.insert(result, GRPC.exporters.static(staticObject))
+      end
+    end
```
justin-lovell, line 132: Implemented as such to keep consistency with behavior in `getGroups`:
https://github.com/DCS-gRPC/rust-server/blob/9c35b22d8ac2022c0afe5d9b1fe0cb9853354b28/lua/DCS-gRPC/methods/coalitions.lua#L223-L243

protos/dcs/coalition/v0/coalition.proto
```diff
@@ -131,6 +135,16 @@ message AddGroupResponse {
...
 }
 
+message GetStaticObjectsRequest {
+  // the coalition which the statics belong to
+  dcs.common.v0.Coalition coalition = 1;
+}
+
+message GetStaticObjectsResponse {
+  // the list of statics
+  repeated dcs.common.v0.Static statics = 1;
```
rurounijones, line 145: Nit: I _think_ that, so far, the return field has matched the request. So `groups` for `getGroups` etc. So maybe this would be better named `static_objects`.

Not too fussed about it though. Hence just a nit.

rurounijones (approved):

PR #149, opened 2022-05-19
Add `GetUnitType` API

MorpheusXAUT:
Tiny PR to add [`DCS.GetUnitType`](https://wiki.hoggitworld.com/view/DCS_func_getUnitType) 🙂 

As discussed via Discord, I originally has this located as under the `UnitService`, where I would instinctively look for it, but since it uses the `DCS` singleton, we have to have it reside within `hooks.lua`. I'm indifferent though about where to keep it - it's in `HookService` now to be consistent with the rest of the hook functions, but I can also move it to `UnitService` if that'd be preferred.

### DIFF
--- a/protos/dcs/hook/v0/hook.proto
+++ b/protos/dcs/hook/v0/hook.proto
@@ -47,6 +47,9 @@ service HookService {
   // Get a list of all the banned players
   rpc GetBannedPlayers(GetBannedPlayersRequest)
     returns (GetBannedPlayersResponse) {}
+
+  // https://wiki.hoggitworld.com/view/DCS_func_getUnitType
+  rpc GetUnitType(GetUnitTypeRequest) returns (GetUnitTypeResponse) {}
 }
 
 message GetMissionNameRequest {
@@ -162,4 +165,14 @@ message BanDetails {
   uint64 banned_from = 5;
   // When the ban will expire in unixtime
   uint64 banned_until = 6;
+}
+
+message GetUnitTypeRequest {
+  // The slot or unit ID of the unit to retrieve the type of
+  string id = 1;
+}
+
+message GetUnitTypeResponse {
+  // Type of unit (e.g. "F-14B")
+  string type = 1;
 }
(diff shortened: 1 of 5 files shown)


rurounijones (approved):

protos/dcs/hook/v0/hook.proto
```diff
@@ -163,3 +166,13 @@ message BanDetails {
   // When the ban will expire in unixtime
   uint64 banned_until = 6;
 }
+
+message GetUnitTypeRequest {
+  // The slot or unit ID of the unit to retrieve the type of
+  string id = 1;
```
rkusa, line 172: This is a number according to https://wiki.hoggitworld.com/view/DCS_func_getUnitType. Not sure if we should care, as it probably also seem to work with strings?

rkusa (commented):
Thanks for the PR 👍

protos/dcs/hook/v0/hook.proto
MorpheusXAUT: Hmm, good question. It definitely works with strings as well, tried that yesterday.
I believe we even have to keep it as string since it works for slot IDs as well and those can have a `_2` suffix for multicrew aircraft, right (e.g. `4_2` for events for an F14 RIO).
Leaving it as string would also be consistent with the [`GetPlayerInfo.Slot`](https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/net/v0/net.proto#L62) (which I based this of).

rkusa (approved):

protos/dcs/hook/v0/hook.proto
rurounijones: The docs are a bit vague with `missionid`. Is this actually the slot ID? The unit ID? Can it be both? Did you try with both?

MorpheusXAUT: I was asking myself the same thing, but yes, it can be both slot ID and unit ID.
I added a single F14 to the example mission provided by you (which gave me the slot ID "2"), then also tried with the E3 (unit ID "1").

Not sure why those are called differently (or even `missionid`), I would simply assume that the slot ID is (roughly) equal to a unit ID?

Anyways, it did work, yes 👍

rurounijones: Cool. Thanks for the confirmation!

PR #150, opened 2022-05-23
Update Lua module methods to never return errors

### LINKED ISSUE #138: Don't return errors from Rust to Lua
It looks like errors returned from Rust to Lua might lead to DCS crashing once a mission is stopped. So either figure out what is wrong with those errors, or don't return errors at all and simply log them with `log::error!` instead.

rkusa:
We had issues with a crash from an error returned by the `dcs_grpc.dll` in the past, see https://github.com/DCS-gRPC/rust-server/issues/19. We fixed that by directly logging the offending error instead of returning it to Lua, see https://github.com/DCS-gRPC/rust-server/commit/bd54607d0d69b93ceed1815ff105058ea06cadb1.

I've investigated this again. I tried to reproduce this issue outside of DCS (basically with a `lua.exe` and a minimal Lua module written in Rust using `mlua`), but was unsuccessful - I couldn't reproduce it outside of DCS. As I don't have any idea left of how to fix/narrow down this issue, I give up and and would suggest to avoid returning errors from Rust to Lua entirely. This is also what this PR changes. It ensures that all calls to `grpc.*` methods never error. If they encounter an error internally, the error is logged right away and not returned to Lua first.

Drawback: The errors will not land in the `Logs/dcs.log` file and instead only in `Logs/gRPC.log`

Potentially fixes the crash reported by the Enigma server admins (see https://discord.com/channels/696721363276267590/962651858760003594/977730972206366781).
Fixes #138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GetTheatre` API
 - `GetUnitType` API
 
+### Changed
+- Improved reliability of error handling.
+
 ## [0.5.0] - 2022-04-19
 ### Added
 - `GetMissionFilename` API
--- a/src/hot_reload.rs
+++ b/src/hot_reload.rs
@@ -1,8 +1,8 @@
-use std::path::PathBuf;
 ///! This module is a wrapper around all exposed Lua methods which are forwarded to a dynamically
 ///! loaded dcs_grpc.dll. Upon calling the `stop()` method, the library is unloaded, and re-
 ///! loaded during the next `start()` call.
-use std::sync::{Arc, RwLock};
+use std::path::PathBuf;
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GetTheatre` API
 - `GetUnitType` API
 
+### Changed
+- Improved reliability of error handling.
```
rurounijones, line 20: Nitpick. Could we expand on this where we mention that all errors are now logged only to `grpc.log`

rurounijones (approved):

rkusa:
On hold for now, see https://github.com/DCS-gRPC/rust-server/issues/138#issuecomment-1146727810

rurounijones:
Closing as no longer needed.

PR #151, opened 2022-05-23
Fix event handler error message being nil

rkusa:
See inline comment

### DIFF
--- a/lua/DCS-gRPC/grpc.lua
+++ b/lua/DCS-gRPC/grpc.lua
@@ -223,7 +223,7 @@ if isMissionEnv then
     local _ = self -- make linter happy
 
     if not stopped then
-      local ok, result, err = xpcall(function() return GRPC.onDcsEvent(event) end, debug.traceback)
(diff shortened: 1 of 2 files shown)


lua/DCS-gRPC/grpc.lua
```diff
@@ -223,7 +223,7 @@ if isMissionEnv then
     local _ = self -- make linter happy
 
     if not stopped then
-      local ok, result, err = xpcall(function() return GRPC.onDcsEvent(event) end, debug.traceback)
+      local ok, result = xpcall(function() return GRPC.onDcsEvent(event) end, debug.traceback)
```
rkusa, line 226: I was using `xpcall` wrong here, which caused the logged error message to be `nil`:
```
[GRPC] Error in event handler: nil
```

rurounijones: Not massivel up to speed on this area of lua but based on http://www.gammon.com.au/scripts/doc.php?lua=xpcall this fix looks corect.

rurounijones (approved):

PR #153, opened 2022-05-27
Add Mission Management APIs

rurounijones:
Add mission management APIs that allow clients to load and restart
missions. The `LoadNextMission` API is apparently supposed to return
false when the last mission in the mission list has been reached but DCS
is returning true all the time.

The API to reload the current mission is not natively provided
by DCS but is simple to implement and will be useful for admins I think.

### DIFF
--- a/protos/dcs/hook/v0/hook.proto
+++ b/protos/dcs/hook/v0/hook.proto
@@ -25,6 +25,21 @@ service HookService {
   // https://wiki.hoggitworld.com/view/DCS_func_stopMission
   rpc StopMission(StopMissionRequest) returns (StopMissionResponse) {}
 
+  // Reload the currently running mission
+  rpc ReloadCurrentMission(ReloadCurrentMissionRequest)
+    returns (ReloadCurrentMissionResponse) {}
+
+  // Load the next mission in the server mission list. Note that it does
+  // not loop back to the first mission once the end of the mission list
+  // has been reached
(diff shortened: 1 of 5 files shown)


protos/dcs/hook/v0/hook.proto
```diff
@@ -87,6 +102,30 @@ message SetPausedRequest {
...
+message LoadNextMissionResponse {
+  // Was the next mission successfully loaded. SHOULD return false when the
+  // end of the mission list has been reached but DCS appears to always
+  // return true
+  bool loaded = 1;
+}
+
+message LoadMissionRequest {
+  // The full path to the .miz file to be loaded
+  string file_name = 1;
```
justin-lovell, line 123: Shall we rename this to `file_path` as this is the full path to where the `.miz` is located?

justin-lovell (approved):
Minor suggestion; otherwise, I will be using these functions in the near future :)

protos/dcs/hook/v0/hook.proto
rurounijones: Yeeaaah, I was never convinced about this naming. I stuck with it since that is what ED uses and I was wondering if we would want to limit the path to the Mission dir or something to avoid potential issues with loading files it shouldn't be.

rurounijones: Deferred to https://github.com/DCS-gRPC/rust-server/issues/154 . For `0.6.0` we will use the above as-is.

Issue #143, opened 2022-05-06
GetSessionId API


rurounijones:
Since DCS-gRPC clients have a lifecycle outside of the DCS server we need a way for them to know if a mission has restarted or a new one loaded. At the moment the only thing a client will know is that it got disconnected and then reconnected.

After bouncing some ideas around with @rkusa I think what we want is:

1. Implement a `Mission.GetSessionId` (Namespace up for debate) `DCS-gRPC` method.
2. When called it checks for the existance of a DCS userFlag `dcs_grpc_session_id`
3. If no value exists then set the value to the current unixtime (since flags can only contain bool or numeric) using something like `os.time(os.date("!*t"))` and return the value.
4. If a value exists then return the existing value.

This will generate a new session unique ID each time a mission is loaded and return the same value until another one is loaded. Clients can then use this ID to determine if a session has changed after getting disconnected and perform appropriate processing.

Documenting this here in case anyone wants to give it a go, otherwise I will get around to it at some point. This would be a good, simple, API to implement for a newcomer.

rkusa:
> 3. If no value exists then set the value to the current unixtime (since flags can only contain bool or numeric) using something like `os.time(os.date("!*t"))` and return the value.

Alternatively, the unixtime could also be retrieved inside of the Rust part of the implementation via [OffsetDateTime::unix_timestamp](https://docs.rs/time/latest/time/struct.OffsetDateTime.html#method.unix_timestamp). This way, there wouldn't be any additional Lua part necessary for this feature, as one could simply re-use the existing lua methods for `GetUserFlag`/`SetUserFlag` from Rust.

justin-lovell:
The selection issue of the `UserFlag` number might be a problem (never know about a conflict?)

You could probably just store this as a start-up variable (initialized `OffsetDateTime::unix_timestamp`) within Rust and just return it directly? No need to jump down into the mission environment.

If the mission ends and starts a new one, reset could be triggered with `handler.onMissionLoadEnd()` chain.

rurounijones:
Implemented by @vivanov

PR #158, opened 2022-06-13
Add GetSessionId API

vivanov:
An attempt to implement following issue: https://github.com/DCS-gRPC/rust-server/issues/143

### DIFF
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -76,6 +76,11 @@ service MissionService {
   // See https://wiki.hoggitworld.com/view/DCS_func_removeItemForGroup
   rpc RemoveGroupCommandItem(RemoveGroupCommandItemRequest)
       returns (RemoveGroupCommandItemResponse) {}
+
+  // Returns an ID for the current session.
+  // The ID will change upon mission change or server restart.
+  rpc GetSessionId(GetSessionIdRequest)
+      returns (GetSessionIdResponse) {}
 }
 
 message StreamEventsRequest {
@@ -719,4 +724,11 @@ message RemoveGroupCommandItemRequest {
 }
 
 message RemoveGroupCommandItemResponse {
+}
+
+message GetSessionIdRequest {
+}
+
+message GetSessionIdResponse {
+  int64 session_id = 1;
 }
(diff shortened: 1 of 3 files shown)


src/rpc/mission.rs
```diff
@@ -165,6 +166,35 @@ impl MissionService for MissionRpc {
...
+        let response = self
+            .get_user_flag(Request::new(trigger::v0::GetUserFlagRequest {
+                flag: String::from("dcs_grpc_session_id"),
+            }))
+            .await?
+            .into_inner();
+        let current_session_id = response.value as i64;
+        if current_session_id == 0 {
+            let ts = OffsetDateTime::now_utc().unix_timestamp();
+            let _result = self
```
rkusa, line 183: I think the following can be removed
```suggestion
            self
```

protos/dcs/mission/v0/mission.proto
```diff
@@ -76,6 +76,10 @@ service MissionService {
   // See https://wiki.hoggitworld.com/view/DCS_func_removeItemForGroup
   rpc RemoveGroupCommandItem(RemoveGroupCommandItemRequest)
       returns (RemoveGroupCommandItemResponse) {}
+
+  // Returns value of the dcs_grpc_session_id flag or current unixtime if it's not set
```
rkusa, line 80: Suggestion (basically fewer implementation details):
```suggestion
  // Returns an ID for the current session. The ID will change upon mission change or server restart.
```

rkusa (commented):
Thanks for the PR, just two minor suggestions 👍  (and protolint wants something: https://github.com/DCS-gRPC/rust-server/runs/6860585828?check_suite_focus=true)

src/rpc/mission.rs
vivanov: Good point, updated

protos/dcs/mission/v0/mission.proto
vivanov: Good point, updated

rurounijones (changes requested):
@vivanov Thanks for the contribution and sorry for the delay in testing. As far as I can tell everything is hunky-dory so thank you very much.

Could you update the CHANGELOG.md to include your change and then I will merge

rurounijones (approved):

Issue #159, opened 2022-06-13
Unit exporter `heading` is not actually units heading, it's the units 2d velocity vector


bitbubbler:
https://github.com/DCS-gRPC/rust-server/blob/main/lua/DCS-gRPC/exporters/object.lua#L28-L30

The `heading` returned for a unit is only correct if the unit is moving exactly forward. This is because `getVelocity()` is being used to calculate the "heading". The result is actually a 2d "velocity vector", which is the correct velocity vector. If the unit is moving forward it should match or be very close to the units heading.

However, it is not the units heading. A unit moving backwards such as a helicopter will result in "heading" from the unit exporter equal to the direction the unit is moving.

The correct way to get the units heading appears to be with the `unit:getPosition()` function [(hoggit wiki)](https://wiki.hoggitworld.com/view/DCS_func_getPosition) using the same algo in use.

This is supported by this forum thread: https://forum.dcs.world/topic/148896-converting-a-units-orientation-to-a-quotheadingquot-script-help/

I'm able to confirm this with `CustomService.Eval` 
(I'm working in typescript, you can do the same in any language. Below code is a snippet)
```typescript
const lua = `return Unit.getByName("${name}"):getPosition()`

custom.eval({ lua }, (error, result) => {
  if (error) {
    console.log('failed to get unit position', error)
    return
  }
  const pos = JSON.parse(result!.json!)

  let heading = deg(Math.atan2(pos.x.z, pos.x.x))
  if (heading < 0) {
    heading = heading + 360
  }

  // here, heading value is always correct, even when moving backwards
})

function deg(radians: number): number {
  return radians * (180 / Math.PI)
}
```

warning: opinions below:

As a consumer of dcs-grpc, I want the units actual heading, In addition to the units velocity. (speed, velocity vector, etc)

I propose that instead of (or in addition to) extracting the heading and speed, expose the units raw velocity vec3. In addition to the units position vector 

There seems to be some attempt to expose the data available in lua outside of lua, but for units it doesn't align with the lua stuff very well. I had to figure out what `heading` was here. And currently have no access to raw vecs to do my own math

rkusa:
Thanks for the report! I haven't really thought about it before, but now that I do, I'd personally expect the heading to be the direction a unit is moving to regardless of where the nose is pointing to.

> As a consumer of dcs-grpc, I want the units actual heading, In addition to the units velocity. (speed, velocity vector, etc)
> 
[...]

PR #162, opened 2022-06-19
Include transform in each unit export

rkusa:
Opening as draft to see if this is something we want to do. If so, I'll finish it up (fixing tests, adding missing changelog entry, etc).

This PR includes the information currently only retrieved via `GetTransform` in the `Unit` message. This makes this data available everywhere we have units (unit stream, events, ...).

## Motivation

I am still working on an LSO, which was also the reason I've added `GetTransform` in the first place. I've now faced the challenge that I needed the information of `GetTransform` at the exact time of a land event. Sending a `GetTransform` after receiving the event yields data that is already newer than the event (units already moved since), which is why I need the information included in the event itself.

With this change, the information also ends up in the units stream, which would solve parts of #159.

## Drawback

More calculations and Lua calls done per unit export. Might effect the performance - though I am not sure if it is even a realistic effect to worry about?

### DIFF
--- a/lua/DCS-gRPC/exporters/object.lua
+++ b/lua/DCS-gRPC/exporters/object.lua
@@ -5,48 +5,42 @@
 -- be kept in sync
 --
 
--- Convert DCS's unusual right-hand coordinate system where +x points north to a more common
--- left-hand coordinate system where +z points north (and +x points east).
-GRPC.exporters.vector = function(v)
-  return {
-    x = v.z,
-    y = v.y,
-    z = v.x
-  }
-end
-
 GRPC.exporters.position = function(pos)
   local lat, lon, alt = coord.LOtoLL(pos)
   return {
     lat = lat,
     lon = lon,
     alt = alt,
+    u = pos.z,
+    v = pos.x,
   }
 end
 
 GRPC.exporters.unit = function(unit)
-  local vector = unit:getVelocity()
-
----@diagnostic disable-next-line: deprecated
-  local heading = math.deg(math.atan2(vector.z, vector.x))
-  if heading < 0 then
-    heading = heading + 360
-  end
-
-  local speed = math.sqrt((vector.x)^2+(vector.z)^2)
-
   return {
     id = tonumber(unit:getID()),
     name = unit:getName(),
     callsign = unit:getCallsign(),
     coalition = unit:getCoalition() + 1, -- Increment for non zero-indexed gRPC enum
     type = unit:getTypeName(),
-    position = GRPC.exporters.position(unit:getPoint()),
     playerName = Unit.getPlayerName(unit),
     group = GRPC.exporters.group(Unit.getGroup(unit)),
     numberInGroup = unit:getNumber(),
-    heading = heading,
-    speed = speed,
+    rawTransform = GRPC.exporters.rawTransform(unit),
+  }
+end
+
+-- Data used to calculate position/orientation/velocity on the Rust side.
+GRPC.exporters.rawTransform = function(object)
+  local p = object:getPosition()
+  local position = GRPC.exporters.position(p.p)
+  return {
+    position = position,
+    positionNorth = coord.LLtoLO(position.lat + 1, position.lon),
+    forward = p.x,
+    right = p.z,
+    up = p.y,
+    velocity = object:getVelocity(),
   }
 end
 
@@ -87,8 +81,9 @@ GRPC.exporters.airbase = function(airbase)
     position = GRPC.exporters.position(airbase:getPoint())
   }
 
-  if airbase:getUnit() then
-    a.id = tonumber(airbase:getUnit():getID())
+  local unit = airbase:getUnit()
+  if unit then
+    a.unit = GRPC.exporters.unit(unit)
   end
 
   return a
--- a/protos/dcs/common/v0/common.proto
+++ b/protos/dcs/common/v0/common.proto
@@ -171,6 +171,10 @@ message Position {
   double lon = 2;
   // Altitude in Meters above Mean Sea Level (MSL)
   double alt = 3;
+  // Distance between DCS' map origin to object in meters on west-east axis.
+  double u = 4;
+  // Distance between DCS' map origin to object in meters on north-south axis.
+  double v = 5;
 }
 
 /**
@@ -203,18 +207,17 @@ message Unit {
   string type = 5;
   // The position of the unit
   Position position = 6;
+  // The orientation of the unit in both 2D and 3D space
+  Orientation orientation = 7;
+  // The velocity of the unit in both 2D and 3D space
+  Velocity velocity = 8;
   // The name of the player if one is in control of the unit
-  optional string player_name = 7;
+  optional string player_name = 9;
   // The group that the unit belongs to
-  Group group = 8;
+  Group group = 10;
   // The number of this unit in the group. Does not change as units are
   // destroyed
-  uint32 number_in_group = 9;
-  // The horizontal speed of the unit. If it is doing mach one straight up then
-  // the speed will be 0
-  double speed = 10;
-  // The heading of the unit
-  double heading = 11;
+  uint32 number_in_group = 11;
 }
 
 /**
@@ -291,9 +294,9 @@ message Scenery {
  *
  */
 message Airbase {
-  // The DCS generated ID. Only used when the airfield is also a unit
-  // (e.g. an Aicraft Carrier)
-  optional uint32 id = 1;
+  // Information about the unit, if the airbase is one (e.g. in case of a
+  // carrier).
+  optional Unit unit = 1;
   // TODO: Fill this in
   string name = 2;
   // TODO: Fill this in
@@ -378,15 +381,50 @@ message MarkPanel {
 }
 
 /**
- * A vector in a left-handed coordinate system with +z being north, -z south, +x
- * east, -x west, +y up and -y down (as opposed to DCS' unusual right-handed
- * coordinate system where x is north/ south and z is west/east; the underlying
- * conversion is basically swapping `x` and `z` as in `x=z` and `z=x`).
+ * A vector in a right-handed coordinate system where +x is north, -x south, +z
+ * is east, -z west, +y up and -y down.
  */
 message Vector {
   double x = 1;
   double y = 2;
   double z = 3;
+}
+
+/**
+ * The orientation of an object in 3D space.
+ */
+message Orientation {
+  // The heading the nose of the object points to on a flat world.
+  double heading = 1;
+  // Yaw in degrees - clockwise relative to the true north (this is similar to
+  // the heading, just corrected by the projection error when going from a flat
+  // to a spherical world).
+  double yaw = 2;
+  // Pitch in degrees - positive when taking-off.
+  double pitch = 3;
+  // Roll in degrees - positive when rolling the aircraft to the right.
+  double roll = 4;
+  // The normalized direction the object is pointing to.
+  Vector forward = 5;
+  // The normalized direction the three line (right wing) is pointing to.
+  Vector right = 6;
+  // The normalized up vector (orthogonal to forward and right).
+  Vector up = 7;
+}
+
+/**
+ * The orientation of an object in 3D space.
+ */
+message Velocity {
+  // The heading the object is moving to (use `orientation.heading` to get the
+  // heading the nose is pointing to).
+  double heading = 1;
+  // The horizontal speed of the unit. If it is doing mach one straight up then
+  // the speed will be 0
+  double speed = 2;
+  // The direction the object is traveling to, and speed (magnitude of the
+  // vector) the object is traveling with.
+  Vector velocity = 3;
 }
 
 /**
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -538,7 +538,7 @@ message StreamUnitsResponse {
   }
 
   oneof update {
-    // The unit is either new or its position changed.
+    // The unit is either new or its position or attitude changed.
     dcs.common.v0.Unit unit = 1;
 
     // The unit does not exist anymore.
--- a/stubs/src/common.rs
+++ b/stubs/src/common.rs
@@ -1,0 +1,125 @@
+pub mod v0 {
+    use std::ops::Neg;
+
+    tonic::include_proto!("dcs.common.v0");
+
+    #[derive(Default, serde::Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub(crate) struct RawTransform {
+        pub position: Option<Position>,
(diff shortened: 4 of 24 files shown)


lua/DCS-gRPC/exporters/object.lua
```diff
@@ -25,31 +25,61 @@ GRPC.exporters.position = function(pos)
...
+  local right = position.z
+  local up = position.y
+
+  local heading = math.atan2(forward.z, forward.x)
+  local headingDegrees = math.deg(math.atan2(forward.z, forward.x))
+  if headingDegrees < 0 then
+    headingDegrees = headingDegrees + 360
+  end
+
+  local velocity = object:getVelocity()
```
rkusa, line 63: Most of the above calculations where previously done in Rust, moving them into Lua might effect the performance.

protos/dcs/common/v0/common.proto
```diff
@@ -201,22 +201,17 @@ message Unit {
   Coalition coalition = 4;
   // The DCS type-name of the unit. e.g "MiG-29A", "ZSU_57_2" or "Hawk ln"
   string type = 5;
-  // The position of the unit
-  Position position = 6;
+  // Information about the unit in 3D space.
+  Transform transform = 6;
```
rkusa, line 205: I am still unsure if `transform` is a good name, but still don't have a better idea.

protos/dcs/common/v0/common.proto
```diff
@@ -201,22 +201,17 @@ message Unit {
...
   // destroyed
   uint32 number_in_group = 9;
-  // The horizontal speed of the unit. If it is doing mach one straight up then
-  // the speed will be 0
-  double speed = 10;
-  // The heading of the unit
-  double heading = 11;
   // The group category.
-  GroupCategory category = 12;
+  GroupCategory category = 10;
```
rkusa, line 214: This is a breaking change that might effect some users of DCS-gRPC.

protos/dcs/common/v0/common.proto
```diff
@@ -381,3 +375,44 @@ message Vector {
...
+/**
+ * The orientation of an object in 3D space.
+ */
+ message Orientation {
+  // The normalized direction the object is pointing to.
+  dcs.common.v0.Vector forward = 1;
+  // The normalized direction the three line (right wing) is pointing to.
+  dcs.common.v0.Vector right = 2;
+  // The normalized up vector (orthogonal to forward and right).
+  dcs.common.v0.Vector up = 3;
```
rkusa, line 411: Just a reminder, that `Vector` is still not a 1 to 1 mapping to DCS and rather a oppinionated modification of it (see https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/common/v0/common.proto#L373-L383)

lua/DCS-gRPC/exporters/object.lua
rurounijones: Compared to the other stuff that runs in pure lua environments I am willing to take the risk. We can always run it against a complicated mission and see what the server reports.

protos/dcs/common/v0/common.proto
rurounijones: Works for me.

protos/dcs/common/v0/common.proto
rurounijones: Luckily we are in the 0.x.x 😁

protos/dcs/common/v0/common.proto
rurounijones: I do not care personally. @bitbubbler - Opinions?

lua/DCS-gRPC/exporters/object.lua
```diff
@@ -25,31 +25,61 @@ GRPC.exporters.position = function(pos)
...
+      pitch = math.deg(math.asin(forward.y)),
+      yaw = math.deg(heading - projection_error),
+      roll = math.deg(-math.asin(right.y)),
+      forward = GRPC.exporters.vector(forward),
+      right = GRPC.exporters.vector(right),
+      up = GRPC.exporters.vector(up),
+    },
+    velocity = GRPC.exporters.vector(velocity),
+    heading = headingDegrees,
+    speed = math.sqrt((velocity.x)^2+(velocity.z)^2),
```
rurounijones, line 79: Does this really need to be moved into the transform?

I am seeing "Heading" (which way nose is pointing) and "Speed" is the basic information that will probably work for most clients. OverlordBot, for example, will only care about heading and speed.

I see the information in the `Transform` section as being the "Advanced" stuff for those small number of people who actually care about it and want to do their own calculations while other clients can ignore its existance.

rurounijones (commented):
Overall looks good to me.

lua/DCS-gRPC/exporters/object.lua
rkusa: I personally need the `heading` as part of the `GetTransform` request, other than that, I'd be fine with moving heading and speed out of it. If we move them out of it, I'd probably still include them in the `GetTransform` response as follows:
```proto
Transform transform = 1;
double speed = 2;
double heading = 3;
```

protos/dcs/common/v0/common.proto
rkusa: (I'd be fine with just using DCS' coordinate system and map them in my application rather than in DCS-gRPC)

bitbubbler: Is there a good reason to not just expose the dcs system? If one want's to port scripts from lua to something that uses grpc why force them to adopt your opinionated system?
Unless it's solving a problem.

justin-lovell:
I'm leaving a few thoughts on the API shapes which could help define a structure / context definition.

The semantics of gRPC appears to follow the theme of external observations. E.g., from a radar scope and movement context, heading is a velocity. 

In the definition of the "heading" in the other discussion where it is the nose position ("aiming point"), that is a concept which would only present from the internal/pilot instrumentation (e.g. HSI).

Therefore, the definition would have a vast difference depending on the context.

If the API is supposed to be opinionated from an external observer point of view, the API shape may be:

```
// physical position in space -- where a "dot" exists
position = { lat, lng, altitude }

// third party observation from a radar scope context (e.g. Overlord)
velocity = { heading, speed, ... and maybe fpm? ... }

// the physical "own ship" details.
orientation = { heading, roll, pitch }

// mapping for DCS
mapping= { up, forward, right }
```

Since there has been a use of the word "opinionated," there is merit in extracting a contextual definition. The `mapping` is an escape hatch for raw values if someone needs to do funky maths beyond the frequently derived variables (velocity, orientation, position)

protos/dcs/common/v0/common.proto
justin-lovell: I reckon building a frequently calculated variables should be "opinionated." E.g. vectors, position, orientation.

However, limited to the discussion of the `Vector`, I agree with @bitbubbler as there are lots of DCS inspired Lua functions for calculating niche things, and I would be annoyed if I needed to change the vector calculations in the ported code in addition translating to another language (e.g. Lua to C#).

Let's expose those raw values (including the map's origin distance in meters) untouched, and where there is an opinion, it is extended on top of those raw values.

lua/DCS-gRPC/exporters/object.lua
rkusa: Here is another idea. How about completely flattening the `transform` into the `Unit` message, so `unit.transform.velocity` becomes `unit.velocity` and so on. I'd then still tend to keep the `GetTransform` API call to get the subset of properties of the unit that describe it in 2D and 3D space (so basically everything currently in the `transform` message). Wdyt?

protos/dcs/common/v0/common.proto
rkusa: @omwdunkley called it _spatial_ [here](https://github.com/DCS-gRPC/rust-server/issues/159#issuecomment-1170409986), so maybe also a good candidate for a name? As a non-native English speaker I don't know which fits better. I'll keep `transform` unless someone else jumps in promoting a different name.

protos/dcs/common/v0/common.proto
rkusa: I'll update the code to use the raw values, thanks for the feedback everyeone 👍

lua/DCS-gRPC/exporters/object.lua
rkusa: I'd like to redact my idea above. While working again on the PR, I've come to realize that I personally found it really useful that all those information are in one message/struct, as this allows them to be easily reused in different contexts (in the unit, in the dedicated method to get the transform, in the units stream).

protos/dcs/mission/v0/mission.proto
```diff
@@ -538,11 +538,14 @@ message StreamUnitsResponse {
   }
 
   oneof update {
-    // The unit is either new or its position changed.
-    dcs.common.v0.Unit unit = 1;
+    // The unit is new.
+    dcs.common.v0.Unit new = 1;
+
+    // The unit's state changed (position, orientation, ...).
+    dcs.common.v0.Transform changed = 2;
```
rkusa, line 545: By grouping everything related to 2D/3D position/transform into one message, we could easily distinguish in the units stream between new and changed, and skip re-sending static unit information again for the latter. Just in case this is something we want to do - I don't have a strong oppinion here myself, so I'd fine with keeping it as is.

rurounijones: I would much prefer we do not do this. Doing so forces clients to maintain state now which not all clients will want to do. I would prefer we send the complete set of unit information, including static information, each time. I think the performance impact of this would be negligable.

protos/dcs/common/v0/common.proto
```diff
@@ -380,13 +375,52 @@ message MarkPanel {
...
 }
+
+/**
+  * Information about an object in 3D space, including its position,
+  * orientation and velocity.
+  */
+message Transform {
+  dcs.common.v0.Position position = 1;
+  // Distance between DCS' map origin to object in meters on west-east axis.
+  double u = 2;
```
rurounijones, line 394: While I know this was requested I don't actually think it is useful information. I have never seen anyone figure out how to convert the internal coord system to lat/long outside of the DCS lua environment since some of the maps are skewed.

I think it was @justin-lovell who requested this. @justin-lovell, What is the use-case?

protos/dcs/common/v0/common.proto
rurounijones: I think transform works fine.

protos/dcs/common/v0/common.proto
rkusa: I need them when exporting units into a [ACMI (TacView) format](https://www.tacview.net/documentation/acmi/en/). Since DCS is basically just a flat world, I had way better results when using `u`/`v` compared to just using lat/lon.

rkusa:
@justin-lovell Thank you for your input! Some remarks:
- I kinda like the suggestion of moving `heading` and `speed` into `velocity`, and `heading` into `orientation` (as it is technically `yaw`) - though I am not sure if having two `heading` values is to confusing and we should rather keep the one in `orientation` named `yaw`?
- I'd propably prefer to keep `up, forward, right` inside of `orientation`  as it gives it some more context of what the values are (vectors describing the orientation in 3D space).
- For consistency, it would then also make sense to move `u` and `v` into `position`.

This would bring the structure to:

```lua
{
  position = { lat, lng, altitude, u, v },
  orientation = { yaw (or heading), roll, pitch, forward = { x, y, z }, right = { x, y, z }, up = { x, y, z } },
  velocity = { heading, speed, direction = { x, y, z } }, -- not sure if `direction` is a good name here for the vector that describes the velocity
}
```

What I still find a bit confusin is the mix of properties defining the object in 3D space, and the properties defining the object as if the world was flat. Values for the latter are: `position.{u,v}`, `velocity.heading`, `velocity.speed`. 🤔

rkusa:
I've made the following changes:

- Grouped the _transform` into:
  ```lua
  {
    position = { lat, lng, altitude, u, v },
    orientation = { heading, yaw, roll, pitch, forward = { x, y, z }, right = { x, y, z }, up = { x, y, z } },
    velocity = { heading, speed, velocity = { x, y, z } }
  }
  ```
- Added `position`/`orientation`/`velocity` directly to `Unit`, instead of grouping it into `transform` (solves my ongoing unhappiness with the term _transform_)
- Moved the math back into the Rust part of the code

I am now happy with the changes.

protos/dcs/common/v0/common.proto
```diff
@@ -171,6 +171,10 @@ message Position {
   double lon = 2;
   // Altitude in Meters above Mean Sea Level (MSL)
   double alt = 3;
+  // Distance between DCS' map origin to object in meters on west-east axis.
+  double u = 4;
+  // Distance between DCS' map origin to object in meters on north-south axis.
+  double v = 5;
```
rkusa, line 177: This adds `u`/`v` to all positions. I'd be up for changing the implementation to only include it for the position of units (might also be useful in the future for projectiles in case anyone wants to export them to TacView).

stubs/src/common.rs
```diff
@@ -0,0 +1,125 @@
...
+        callsign: String,
+        coalition: i32,
+        r#type: String,
+        player_name: Option<String>,
+        group: Option<Group>,
+        number_in_group: u32,
+        raw_transform: Option<RawTransform>,
+    }
+
+    impl From<UnitIntermediate> for Unit {
```
rkusa, line 96: This code allows to move the math back to Rust and automitcally apply it to all `Unit`s, regardless of where they are exported.

stubs/src/lib.rs
```diff
@@ -2,88 +2,25 @@
...
-        tonic::include_proto!("dcs.unit.v0");
-    }
-}
-
-pub mod world {
-    pub mod v0 {
-        tonic::include_proto!("dcs.world.v0");
-    }
-}
+pub mod world;
```
rkusa, line 18: I've moved all those `mod`s into a separate file. This improved the incremental compilation a lot when making changes to one of them.

protos/dcs/common/v0/common.proto
rurounijones: 👍

stubs/src/common.rs
rurounijones: Sounds like a good decision.

stubs/src/lib.rs
rurounijones: Very nice

rurounijones (approved):
On mobile so checking all the details is beyond at the moment but ddoximentatoon and whatnot are understandable so I think it is good from my end. Not sure if you want someone to have a closer look at the code.

rkusa:
Rebased due to merge conflicts. I'll be going ahead with merging this. I've tested the code in my LSO bot. I am happy to create follow-ups, if anything comes up (code-wise, or who the proto messages are structured).

PR #167, opened 2022-06-28
Add FPS event

### LINKED ISSUE #164: Add server FPS to the event stream
Since DCS-gRPC is hooked into the hook environment and gets called on each simulation frame; we can calculate the FPS of the server and emit that value as an event every second. This will allow clients to be able to see the server FPS which is useful for diagnostic purposes.

One solid use-case is to use `Telemachus` to record FPS metrics for display.

rkusa:
Calculate FPS by hooking into `onSimulationFrame` and fire an event
every second containing the FPS since the last event (~1s).

Resolves #164

image

### DIFF
--- a/protos/dcs/mission/v0/mission.proto
+++ b/protos/dcs/mission/v0/mission.proto
@@ -377,7 +377,7 @@ message StreamEventsResponse {
     // The slot's identifier
     string slot_id = 3;
   }
-  
+
   /**
    * Fired when a player connected to the server.
    */
@@ -443,6 +443,15 @@ message StreamEventsResponse {
     google.protobuf.Struct details = 2;
   }
 
+  /**
+   * Fired every second containing simulation FPS information since the previous
+   * event.
+   */
+  message SimulationFpsEvent {
+    // The average FPS since the last event.
+    double average = 1;
+  }
+
   // The event's mission time.
   double time = 1;
   oneof event {
@@ -494,6 +503,7 @@ message StreamEventsResponse {
     MissionCommandEvent mission_command = 8196;
     CoalitionCommandEvent coalition_command = 8197;
     GroupCommandEvent group_command = 8198;
+    SimulationFpsEvent simulation_fps = 8199;
   }
 }
 
@@ -604,7 +614,7 @@ message RemoveMissionCommandItemResponse {
 // The emitted event will include the coalition.
 message AddCoalitionCommandRequest {
   // The coalition whose players will be able to see and run the command
-  dcs.common.v0.Coalition coalition = 1; 
+  dcs.common.v0.Coalition coalition = 1;
   // The name of the command that is displayed to the player.
   // It will form the last entry in the returned path.
   string name = 2;
@@ -644,7 +654,7 @@ message AddCoalitionCommandSubMenuResponse {
 
 message RemoveCoalitionCommandItemRequest {
   // The coalition whose players will have the menu item removed
-  dcs.common.v0.Coalition coalition = 1; 
+  dcs.common.v0.Coalition coalition = 1;
   // The full path to the menu item, which can be a submenu or a command, to be
   // removed. Deleting a menu item will delete all children it may have.
   repeated string path = 2;
--- a/src/fps.rs
+++ b/src/fps.rs
@@ -1,0 +1,57 @@
+use std::future::Future;
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::time::Duration;
+
+use dcs_module_ipc::IPC;
+use stubs::mission::v0::stream_events_response::{Event, SimulationFpsEvent};
+use stubs::mission::v0::StreamEventsResponse;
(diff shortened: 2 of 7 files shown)


src/fps.rs
```diff
@@ -0,0 +1,59 @@
...
+        ipc.request::<_, stubs::trigger::v0::OutTextResponse>(
+            "outText",
+            Some(stubs::trigger::v0::OutTextRequest {
+                text: format!("Server FPS: {:.2}", fps),
+                display_time: 2,
+                clear_view: true,
+            }),
+        )
+        .await
+        .ok();
```
rkusa, line 57:
```suggestion
```

Needs to be removed before merge. Just in here to ease testing. I've compared its output with DCS' FPS counter and Nividia's performance overlay and seem to match up close enough.

protos/dcs/mission/v0/mission.proto
```diff
@@ -443,6 +443,14 @@ message StreamEventsResponse {
     google.protobuf.Struct details = 2;
   }
 
+  /**
+   * Fired every second containing the FPS since the previous event.
+   */
+  message FpsEvent {
```
rkusa, line 449: Wdyt, should it be called `ServerFpsEvent`?

rkusa:
Todo:
- [ ] handle paused mission

protos/dcs/mission/v0/mission.proto
rurounijones: I think `ServerFpsEvent` is the better name or maybe a more accurate `SimulationFPSEvent`  / `ServerSimulationFPSEvent` (It is a longer name but more accurate and I don't mind long names if there is a reason. However I am fine with any of the names to be honest).

rurounijones:
Out of curiosity; what extra work needs to be done to handle the paused state?

protos/dcs/mission/v0/mission.proto
justin-lovell: I would concur with `SimulationFpsEvent` -- I have kind of started work on a single-player mechanic (so not necessarily it's a server)

justin-lovell:
> Out of curiosity; what extra work needs to be done to handle the paused state?

I am anticipating that the simulation loop does not trigger the counter into the Rust FPS, and therefore the FPS would read as 0 instead of a `null` or similar. 

I would anticipate the Rust code querying the simulation state (`DCS.getPause`) and not emit any per second event through the event stream.

Otherwise, the code is looking good.

protos/dcs/mission/v0/mission.proto
rkusa: I like `SimulationFpsEvent` 👍 - I'll change it in the next update of the PR

rkusa:
DCS is suprisingly still calling `onSimulationFrame` even when it is paused. So right now, the implementation would still emit an FPS event every second, but the `time` of the event isn't advancing. Example:

```json
{
  "time": 16.589,
  "fps": {
    "fps": 63
  }
}
{
  "time": 16.589,
  "fps": {
    "fps": 63
  }
}
{
  "time": 16.589,
  "fps": {
    "fps": 71
  }
}
{
  "time": 16.589,
  "fps": {
    "fps": 72
  }
}
```

I assume we rather want to not fire any FPS event while the simulation is paused? If so, I would base the 1 second interval on the DCS simulation time and not on the real time - ie. it would only emit an event once 1 second of the `time` property has passed.

protos/dcs/mission/v0/mission.proto
```diff
@@ -443,6 +443,14 @@ message StreamEventsResponse {
...
+    // The FPS since the last FpsEvent.
+    double fps = 1;
```
rkusa, line 451: Some additional ideas/thoughts:
1. Should we rename this to `average`?
2. Should we add a `min` and `max` as well? (for me, DCS easily peaks at over 100FPS, but the average for the second still ends up being at about 70 FPS)

rkusa: > I am assuming you mean the max and min is based on the quickest / slowest time between two individual frames in the second?
Yes 👍

rurounijones (commented):
Looks fine to me with the proposed event name change and removing the temporary stuff. Thank you for doing this so quickly.

protos/dcs/mission/v0/mission.proto
rkusa: I have a local branch with the highest/lowest FPS, but didn't include it in the PR yet. I always got something like: 70 average, 60 lowest, and 250 highest. The 250 highest was almost always reported as the highest. It looks like there are always one or a couple of simulation frames that are pretty quick. With that, I am unsure if it is still helpful information. Feel free to let me know if you'd still like it included.

rurounijones: given what you have discovered lets go with the average for the moment; that should still be enough to give server owners insight into what is going on.

rurounijones (approved):
Thank you for getting this done so quickly!

Issue #163, opened 2022-06-27
Return parent objects


rurounijones:
Return parent objects intead of references.

For example, when returning a unit we don't return a `group_name` or `group_id` field. We return the parent group object. We also return the `player` object (If one exists).

This will flesh out the information and allow for more advanced processing by event consumers and reduce the chance of race-conditions.

Note that we will only ever return the parent object. We wil not return children of the parent object since that will result in, at worst, an infinite loop (e.g. we will not return a unit with a group with a list of units, each with a group etc.).

PR #172, opened 2022-08-06
Return the full parent group with the unit

rurounijones:
When returning a unit object we will return the full group as its own
object instead of embedding some group attributes in the unit object.
This makes it easier for clients to parse streaming events without
having to make separate calls to get group information or require them
to keep a local copy in state that then needs to be managed.

Partly resolves https://github.com/DCS-gRPC/rust-server/issues/163


rkusa (approved):
👍

PR #177, opened 2022-09-06
Convert scenery name to a string

soerenmeier:
When listening to Dead events, no events where triggered for scenery objects.
After looking at the logs I found out that the deserialization failed because
the scenery struct expected a string but a number was given.

Since I don't know if all scenery units have an integer as a name my idea
was to convert the integer to a string.

If scenery object names can only be numbers it would probably make more sense
to change the scenery struct to use an `uint`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Unit objects now return the full group object in the `group` field to make event processing easier. This replaces the `group_name` and `group_category` fields and is a backwards incompatible change.
 - Updated all vectors to be in DCS' coordinate system (+x north, -x south, +z is east, -z west, +y up and -y down)
+- Scenery objects now have an `id` instead of a `name`, since dcs associates them with a number.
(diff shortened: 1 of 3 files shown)


rkusa (commented):
Thanks for PR!

> If scenery object names can only be numbers it would probably make more sense to change the scenery struct to use an `uint`.

That is something I'd be curious about before merging the PR. I'll do some tests myself to see if I can find any scenery object with a name of type string.

rkusa:
I think we can change it to a number. I did some samples by searching areas for sceneries using [`searchObjects`](https://wiki.hoggitworld.com/view/DCS_func_searchObjects), and all had a numeric id. If we encounter any string id in the future, we can still change it back to a string and convert the numberic ids accordingly.

soerenmeier:
Will test the new commit tonight, but I think it should work.

rkusa (commented):
Works for me. Could you add an entry to the `CHANGELOG.md`? Otherwise it'd be good to go.

soerenmeier:
Done

rurounijones (approved):
Thank you for the contribution!

PR #183, opened 2022-09-23
Apply cargo workspace inheritance to sub-crates

rkusa:
This PR makes use of [Rust 1.64's workspace inheritance](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds). This allows sub-crates to inherit metadata (version, etc.) and dependencies.

Advantage:
- Easier to ensure all sub-crates use the same version of a common dependency
- By inheriting the version, we can not forgot to bump one of them upon release

Disadvantage:
- Bumps min Rust version to 1.64

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,7 +247,7 @@ dependencies = [
 
 [[package]]
 name = "dcs-grpc-repl"
-version = "0.1.0"
+version = "0.6.0"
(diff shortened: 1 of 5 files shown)


rurounijones (approved):
Very nice!

PR #184, opened 2022-09-27
Add getDetectedTargets method to controller

YoloWingPixie:
Created GRPC.methods.getDetectedTargets which returns target name, target type, visible (to the radar), distance (true if known by the radar), and optionally the unit or weapon object.

Created exporter for detectedTarget.

Added GetDetectedTargets rpc call to the controller service.

Added GetDetectedTargetsRequest and Response message.

controller proto now imports common proto.

Defined common Contact proto message

### DIFF
--- a/lua/DCS-gRPC/methods/controllers.lua
+++ b/lua/DCS-gRPC/methods/controllers.lua
@@ -51,37 +51,37 @@ GRPC.methods.getDetectedTargets = function(params)
   for i, contact in ipairs(targets) do
     local category = Object.getCategory(contact.object)
 
-    contact.id = contact.object.id_
-    contact.target = {}
-
-    if category == 1 then
-      contact.name = Unit.getName(contact.object)
-      contact.type = 1
-      if params.includeObject == true then
-        contact.unit = GRPC.exporters.unit( contact.object )
-      end
-    end
-    if category == 2 then
-      contact.name = Weapon.getName(contact.object)
-      contact.type = 2
-      if params.includeObject == true then
-        contact.weapon = GRPC.exporters.weapon( contact.object )
-      end
-    end
-
-    if contact.type == nil then
-      return GRPC.errorNotFound("Could not find target with id '" .. contact.object.id_ .. "' as a Unit or Weapon")
+    if category == nil then
+      return GRPC.errorNotFound("Could not find target with id '" .. contact.object:getID() .. "'")
     end
 
     local result = {
       distance = contact.distance,
-      name = tostring(contact.name),
-      id = contact.id,
-      type = contact.type,
+      id = contact.object.id_,
       visible = contact.visible,
-      unit = contact.unit,
-      weapon = contact.weapon,
+      target = {}
     }
+
+    --If target is a unit
+    if category == 1 then
+      if params.includeObject == true then
+        result.target.unit = GRPC.exporters.unit( contact.object )
+      else
+        result.target.basicInformation = {}
+        result.target.basicInformation.name = tostring( Unit.getName(contact.object) )
+        result.target.basicInformation.type = 1
+      end
+    end
+    --If target is a weapon
+    if category == 2 then
+      if params.includeObject == true then
+        result.target.weapon = GRPC.exporters.weapon( contact.object )
+      else
+        result.target.basicInformation = {}
+        result.target.basicInformation.name = tostring( Weapon.getName(contact.object) )
+        result.target.basicInformation.type = 2
+      end
+    end
 
     results[i] = result
   end
--- a/lua/DCS-gRPC/methods/controllers.lua
+++ b/lua/DCS-gRPC/methods/controllers.lua
@@ -49,21 +49,41 @@ GRPC.methods.getDetectedTargets = function(params)
   local results = {}
 
   for i, contact in ipairs(targets) do
-    contact.name = Unit.getName({["id_"] = contact.object.id_})
+    local category = Object.getCategory(contact.object)
 
-    local contact_unit = Unit.getByName(contact.name)
+    contact.id = contact.object.id_
+    contact.target = {}
 
-    local descriptor
-    if params.includeDescriptors then
-      descriptor = GRPC.methods.getUnitDescriptor({ ["name"] = contact.name }).result.attributes
+    if category == 1 then
+      contact.name = Unit.getName(contact.object)
+      contact.type = 1
+      if params.includeObject == true then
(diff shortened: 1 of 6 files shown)


YoloWingPixie:
Example response for 
`{
  "unitName": "EWR",
  "includeObject": true
}`

```{
  "contacts": [
    {
      "name": "Aerial-6-1",
      "id": 16779520,
      "visible": true,
      "distance": true,
      "type": 1,
      "unit": {
        "id": 9,
        "name": "Aerial-6-1",
        "callsign": "Ford11",
        "coalition": "COALITION_BLUE",
        "type": "F-16C_50",
        "position": {
          "lat": 34.54971055583146,
          "lon": 39.02106228310789,
          "alt": 10659.232421875
        },
        "player_name": "New callsign",
        "group": {
          "id": 9,
          "name": "Aerial-6",
          "coalition": "COALITION_BLUE",
          "category": "GROUP_CATEGORY_AIRPLANE"
        },
        "number_in_group": 1,
        "speed": 230.31925030016578,
        "heading": 63.14218647955991
      }
    },
    {
      "name": "16780544",
      "id": 16780544,
      "visible": true,
      "distance": true,
      "type": 2,
      "weapon": {
        "id": 16780544,
        "type": "AGM_88",
        "position": {
          "lat": 34.56112917717297,
          "lon": 39.07106689398359,
          "alt": 14012.8017578125
        }
      }
    }
  ]
}
```

Without `includeObject`

```
{
  "contacts": [
    {
      "name": "Aerial-6-1",
      "id": 16779520,
      "visible": true,
      "distance": true,
      "type": 1
    },
    {
      "name": "16780544",
      "id": 16780544,
      "visible": true,
      "distance": true,
      "type": 2
      }
    }
  ]
}
```

protos/dcs/common/v0/common.proto
```diff
@@ -388,3 +388,26 @@ message Vector {
...
   double z = 3;
 }
+
+/**
+ * An instance of a contact in a DCS AI controller's detection table
+ *
+ * This is a target that the AI controller has detected and is actively tracking
+ *
+ */
+message Contact {
```
rurounijones, line 398: I think a better way of doing this would be to use `oneof`

Then we could have

```proto
message Contact {
  message BasicInformation {
    // The contact's name
    string name = 1;
    // The DCS generated ID
    uint32 id = 2;
    // The type of unit the contact is
    uint32 type = 3;
  }

  oneof details {
    BasicInformation basic_info = 1;
    Unit unit = 2;
    Weapon weapon = 3;
  }
  bool is_visible = 4; // I am not entirely sure what this means. If it is not visible then it would not be in the list, no?
  bool sitance = 5;
}
```

Which removes all duplication

YoloWingPixie: Agreed,

Committed changes to implement this. 

As for the comment about is_visible...it's documented in the DCS API as if the unit can see the target or not, and it doesn't really go into much detail for that. I don't actually fully understand what it represents, but I included it because it was included in getDetectedTargets().

rurounijones: I think it might be useful for targets that are detected via datalink.

lua/DCS-gRPC/methods/controllers.lua
```diff
@@ -30,3 +30,63 @@ GRPC.methods.setAlarmState = function(params)
...
 end
+
+GRPC.methods.getDetectedTargets = function(params)
+  local unit = Unit.getByName(params.unitName)
+  if unit == nil then
+    return GRPC.errorNotFound("Could not find unit with name '" .. params.unitName .. "'")
+  end
+
+  local controller = Unit.getController(unit)
+  local targets = controller:getDetectedTargets()
```
rurounijones, line 41: We should add the ability to specify by which detection method targets should be looked for as documented in https://wiki.hoggitworld.com/view/DCS_func_getDetectedTargets . This will allow us to break up the monolithic "SHOW ME EVERYTHING IN THIS SIMULATION FRAME" call into a series of smaller ones that return smaller target lists that could be executed over multiple frames.. This could help with performance.

We should also aim to include all built-in method parameters in our APIs.

lua/DCS-gRPC/methods/controllers.lua
```diff
@@ -30,3 +30,63 @@ GRPC.methods.setAlarmState = function(params)
...
+
+    local result = {
+      distance = contact.distance,
+      id = contact.object.id_,
+      visible = contact.visible,
+      target = {}
+    }
+
+    --If target is a unit
+    if category == 1 then
```
rurounijones, line 66: Another option would be to use https://github.com/DCS-gRPC/rust-server/blob/main/lua/DCS-gRPC/exporters/object.lua#L111 

If details not needed. Export using the "Unknown" exporter type.

Otherwise use this if statement.

YoloWingPixie: Odd problem with using the `Unknown` exporter. It's returning a type of `0`:

```    
    {
      "id": 16778240,
      "visible": true,
      "distance": true,
      "object": {
        "name": "Aerial-1-1",
        "type": 0
      }
```

But neither the exporter method nor the Unknown common proto has a type field. This is not causing an error.

```
message Unknown {
  string name = 1;
}
```

```lua
GRPC.exporters.unknown = function(object)
  return {
    name = object:getName(),
  }
end
```

So I can leave this in but it's technically inaccurate or misleading.

lua/DCS-gRPC/methods/controllers.lua
YoloWingPixie: Implemented

protos/dcs/common/v0/common.proto
```diff
@@ -375,7 +375,7 @@ message MarkPanel {
   optional string text = 6;
   // The position of the mark.
   Position position = 7;
-}
+}
```
rurounijones, line 378: rogue whitespace

lua/DCS-gRPC/methods/controllers.lua
rurounijones: i don't follow with innacurate since there is no type in the RPC response which I think is fine?

rurounijones (changes requested):
Thanks for the work! Could you update the CHANGELOG.md file, add your change to the list and bask in the glory of contribution.

YoloWingPixie (commented):
Done @rurounijones

rurounijones (approved):

PR #186, opened 2022-10-03
Change Weapon export to include speed and heading

YoloWingPixie:
Changed object.lua to export speed and heading for weapon.

Added heading and speed fields to Weapon proto message.

### DIFF
--- a/lua/DCS-gRPC/exporters/object.lua
+++ b/lua/DCS-gRPC/exporters/object.lua
@@ -5,48 +5,42 @@
 -- be kept in sync
 --
 
--- Convert DCS's unusual right-hand coordinate system where +x points north to a more common
--- left-hand coordinate system where +z points north (and +x points east).
-GRPC.exporters.vector = function(v)
-  return {
-    x = v.z,
-    y = v.y,
-    z = v.x
-  }
-end
-
 GRPC.exporters.position = function(pos)
   local lat, lon, alt = coord.LOtoLL(pos)
   return {
     lat = lat,
     lon = lon,
     alt = alt,
+    u = pos.z,
+    v = pos.x,
   }
 end
 
 GRPC.exporters.unit = function(unit)
-  local vector = unit:getVelocity()
-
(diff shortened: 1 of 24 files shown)


lua/DCS-gRPC/exporters/object.lua
```diff
@@ -60,10 +60,22 @@ GRPC.exporters.group = function(group)
 end
 
 GRPC.exporters.weapon = function(weapon)
+  local vector = weapon:getVelocity()
+
+---@diagnostic disable-next-line: deprecated
+  local heading = math.deg(math.atan2(vector.z, vector.x))
+  if heading < 0 then
+    heading = heading + 360
+  end
```
rkusa, line 69: Probably not as relevant for weapons as for units, but for consistency sake, we might wanna make it somehow clear that this is the heading the weapons travels to, and not the heading the nose of the weapons points to (see #159).

A possible solution would be to copy what #162 does and [group that into a `Velocity` message](https://github.com/DCS-gRPC/rust-server/pull/162/files#diff-488937a422e8fd84be71409cec6dbea6a8dfbe117e43a026ec564b34c807b8c3R418).

rurounijones: I was about to ask if we can add the transform stuff from #162 to weapon exporter as well.

YoloWingPixie: I'm happy to help include the changes, although I think that personally I'd rather know about the ground track of a weapon rather than where its nose is pointing. In a very strong winds aloft, a HARM may have to yaw into the wind by a not insignificant amount of degrees, that if you returned its bearing of its nose as it's heading, you might be mislead on which unit(s) that weapon is targeting.

Would it be possible to include the changes from #162 and include the current heading as `ground_track`?

rurounijones: The changes in #162 includes the nose-pointing heading and the velocity heading

PR #191, opened 2022-10-10
Add time to units stream

### LINKED ISSUE #165: Add mission time to the `StreamUnitsResponse`
Add the timestamp to the [StreamUnitsResponse](https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/mission/v0/mission.proto#L519=) so that clients know exactly when the update occurred. This will bring it inline with the [EventStreamResponse](https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/mission/v0/mission.proto#L446=) which already has the time.

The use-case provided for this was to be able to replay event streams offline for testing and other purposes.

rkusa:
Added DCS `time` of the update to units stream (`StreamUnitsResponse`).

Resolves #165


rurounijones (approved):
Hurrah!

PR #193, opened 2022-10-18
Add SRS client

rkusa:
Add SRS client. Moved over [from DATIS](https://github.com/rkusa/DATIS/tree/main/crates/srs) with minor modifications.

PR #192, opened 2022-10-18
Add Azure, AWS, GCloud & Windows as TTS providers

rkusa:
This is the first in an upcoming PR chain towards adding TTS to DCS-gRPC. It doesn't expose any functionality yet. It just adds Windows and the big three clouds as TTS providers, which will be used in follow-up PRs. I am going to split the PRs up to ease the review.

The code of this PR is mostly just moved over [from DATIS](https://github.com/rkusa/DATIS/tree/main/crates/srs).

No review needed yet.

PR #194, opened 2022-10-18
Implement TTS service

rkusa:
Add TTS service, which allows to synthesize text and transmit it over SRS.

Example:

```bash
.\grpcurl -plaintext -import-path ./protos -proto ./protos/dcs/dcs.proto -d '{\"text_ssml\": \"Hello Pilot!\", \"frequency\": 251000000}' 127.0.0.1:50051 dcs.tts.v0.TtsService/Transmit
```

AWS:

```bash
.\grpcurl -plaintext -import-path ./protos -proto ./protos/dcs/dcs.proto -d '{\"text_ssml\": \"Hello Pilot!\", \"frequency\": 251000000, \"aws\": {}}' 127.0.0.1:50051 dcs.tts.v0.TtsService/Transmit
```

Azure:

```bash
.\grpcurl -plaintext -import-path ./protos -proto ./protos/dcs/dcs.proto -d '{\"text_ssml\": \"Hello Pilot!\", \"frequency\": 251000000, \"azure\": {}}' 127.0.0.1:50051 dcs.tts.v0.TtsService/Transmit
```

Google Cloud:

```bash
.\grpcurl -plaintext -import-path ./protos -proto ./protos/dcs/dcs.proto -d '{\"text_ssml\": \"Hello Pilot!\", \"frequency\": 251000000, \"gcloud\": {}}' 127.0.0.1:50051 dcs.tts.v0.TtsService/Transmit
```

This PR is based on #193 and #192.

TODO:
- [x] Test Win
- [x] Test AWS
- [x] Test Azure
- [x] Test GCloud

### DIFF
--- a/protos/dcs/tts/v0/tts.proto
+++ b/protos/dcs/tts/v0/tts.proto
@@ -30,7 +30,7 @@ message TransmitRequest {
   uint64 frequency = 3;
 
   // Name of the SRS client. Defaults to "DCS-gRPC".
-  optional string name = 4;
+  optional string srs_client_name = 4;
 
   // The origin of the transmission. Relevant if the SRS server has "Line of
   // Sight" and/or "Distance Limit" enabled.
--- a/protos/dcs/tts/v0/tts.proto
+++ b/protos/dcs/tts/v0/tts.proto
@@ -5,6 +5,12 @@ option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Tts";
 option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/tts";
 
 service TtsService {
+  // Synthesize text to speech and transmit it over SRS. By default, this blocks until a
+  // transmission completed (unless `async` is set to `true`). This can be used to prevent
+  // transmission to overlap each other, by not sending another transmission on the same frequency
+  // until you've received the response from the previous transmission on that frequency. However,
+  // it does not block or prevent any other client from transmitting over the same frequency at the
+  // same time.
   rpc Transmit(TransmitRequest) returns (TransmitResponse) {}
 }
 
--- a/protos/dcs/tts/v0/tts.proto
+++ b/protos/dcs/tts/v0/tts.proto
@@ -9,28 +9,35 @@ service TtsService {
 }
 
 message TransmitRequest {
-  // The text that is synthesized to speech and transmitted to SRS.
-  string text = 1;
+  // The text that is synthesized to speech and transmitted to SRS. Supports SSML tags (you should
+  // not wrap the text in the root `<speak>` tag though).
+  string text_ssml = 1;
+
+  // The plain text without any transformations made to it for the purpose of getting it spoken out
+  // as desired (no SSML tags, no FOUR NINER instead of 49, ...). Even though this field is
+  // optional, please consider providing it as it can be used to display the spoken text to players
+  // with hearing impairments.
+  optional string text_plain = 2;
 
   // The radio frequency in Hz the transmission is send to. Example: 251000000
   // for 251.00MHz.
-  uint64 frequency = 2;
+  uint64 frequency = 3;
 
   // Name of the SRS client. Defaults to "DCS-gRPC".
-  optional string name = 3;
+  optional string name = 4;
 
   // The origin of the transmission. Relevant if the SRS server has "Line of
   // Sight" and/or "Distance Limit" enabled.
-  dcs.common.v0.Position position = 4;
+  dcs.common.v0.Position position = 5;
 
   // The coalition of the transmission. Relevant if the SRS server has "Secure
   // Coalition Radios" enabled. Only Blue and Red are supported, all other
   // values will fallback to Blue.
-  dcs.common.v0.Coalition coalition = 5;
+  dcs.common.v0.Coalition coalition = 6;
 
   // The socket (hostname:port) of the SRS server to connect to. Defaults to
   // 127.0.0.1:5002 if not specified.
-  optional string srs_addr = 6;
+  optional string srs_addr = 7;
 
   // Whether to keep the request open until the whole transmission was sent. If
   // enabled, you can send the next transmission after you've received the
@@ -38,7 +45,7 @@ message TransmitRequest {
   // over each other). If disabled, you'll receive a response right away (kind
   // of fire and forget). You can use the returned duration as a spacing between
   // TTS requests to prevent the overlap of multiple playbacks yourself.
-  bool async = 7;
+  bool async = 8;
 
   message Aws {
     // The voice the text is synthesized in, see:
@@ -66,10 +73,10 @@ message TransmitRequest {
   // Optional TTS provider to be use. Defaults to the one configured in your
   // config or to Windows' built-in TTS.
   oneof provider {
-    Aws aws = 8;
-    Azure azure = 9;
-    GCloud gcloud = 10;
-    Windows win = 11;
+    Aws aws = 9;
+    Azure azure = 10;
+    GCloud gcloud = 11;
+    Windows win = 12;
   }
 }
 
--- a/protos/dcs/tts/v0/tts.proto
+++ b/protos/dcs/tts/v0/tts.proto
@@ -1,0 +1,94 @@
+syntax = "proto3";
+package dcs.tts.v0;
+import "dcs/common/v0/common.proto";
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Tts";
+option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/tts";
+
+service TtsService {
+  rpc Transmit(TransmitRequest) returns (TransmitResponse) {}
+}
+
+message TransmitRequest {
+  // The text that is synthesized to speech and transmitted to SRS.
+  string text = 1;
+
+  // The radio frequency in Hz the transmission is send to. Example: 251000000
+  // for 251.00MHz.
+  uint64 frequency = 2;
+
+  // Name of the SRS client. Defaults to "DCS-gRPC".
+  optional string name = 3;
+
+  // The origin of the transmission. Relevant if the SRS server has "Line of
+  // Sight" and/or "Distance Limit" enabled.
+  dcs.common.v0.Position position = 4;
+
+  // The coalition of the transmission. Relevant if the SRS server has "Secure
+  // Coalition Radios" enabled. Only Blue and Red are supported, all other
+  // values will fallback to Blue.
+  dcs.common.v0.Coalition coalition = 5;
+
+  // The socket (hostname:port) of the SRS server to connect to. Defaults to
+  // 127.0.0.1:5002 if not specified.
+  optional string srs_addr = 6;
+
+  // Whether to keep the request open until the whole transmission was sent. If
+  // enabled, you can send the next transmission after you've received the
+  // response for the previous one and be sure that they don't overlap (talk
+  // over each other). If disabled, you'll receive a response right away (kind
+  // of fire and forget).
+  bool wait = 7;
+
+  message Aws {
+    // The voice the text is synthesized in, see:
+    // https://docs.aws.amazon.com/polly/latest/dg/voicelist.html
+    optional string voice = 1;
+
+    // Your AWS API access key.
+    string key = 2;
+
+    // Your AWS API secret key.
+    string secret = 3;
+
+    // The AWS region identifier, see:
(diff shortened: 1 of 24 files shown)


protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+package dcs.tts.v0;
+import "dcs/common/v0/common.proto";
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Tts";
+option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/tts";
+
+service TtsService {
+  rpc Transmit(TransmitRequest) returns (TransmitResponse) {}
+}
+
+message TransmitRequest {
```
rkusa, line 11: While the PR isn't ready for review yet, @rurounijones feel free to give the proto definition a look already (we discussed a few things in voice a while ago, but I unfortunately don't remember everything we agreed on)

src/rpc/tts.rs
```diff
@@ -0,0 +1,159 @@
...
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
+        } else {
+            let signal = self.shutdown_signal.signal();
+            tokio::task::spawn(async move {
+                if let Err(err) = transmit(&request.text, &config, stream, signal).await {
+                    log::error!("TTS transmission failed: {}", err);
+                }
+            });
+        }
```
rkusa, line 104: Future improvement: keep SRS client connections open for a bit to re-use them for subsequent requests

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+    string secret = 3;
+
+    // The AWS region identifier, see:
+    // https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints
+    string region = 4;
+  }
+
+  message Azure {
+    // The voice the text is synthesized in, see:
+    // https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support?tabs=stt-tts
```
rkusa, line 60: Well, protolint isn't happy about those long lines, but I don't want to split the URLs either - any idea?

rurounijones: We could use a URL shortener if there is no way to tell the linter to ignore this line.

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+  // The socket (hostname:port) of the SRS server to connect to. Defaults to
+  // 127.0.0.1:5002 if not specified.
+  optional string srs_addr = 6;
+
+  // Whether to keep the request open until the whole transmission was sent. If
+  // enabled, you can send the next transmission after you've received the
+  // response for the previous one and be sure that they don't overlap (talk
+  // over each other). If disabled, you'll receive a response right away (kind
+  // of fire and forget).
+  bool wait = 7;
```
rurounijones, line 40: Would `async` be a better name for this? Or `blocking`? Both of these are industry standard terms which might make it more obvious.

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+    // The voice the text is synthesized in.
+    optional string voice = 1;
+  }
+
+  oneof provider {
+    Aws aws = 8;
+    Azure azure = 9;
+    GCloud gcloud = 10;
+    Windows win = 11;
+  }
```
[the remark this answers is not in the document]
rurounijones: I could see arguments for both bu I think, for the moment at least, having them be in the config file is the better option. We could add them to the request later and have a "use request options if they exist or fallback to config"

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+
+    // The AWS region identifier, see:
+    // https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints
+    string region = 4;
+  }
+
+  message Azure {
+    // The voice the text is synthesized in, see:
+    // https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support?tabs=stt-tts
+    optional string voice = 1;
```
rurounijones, line 61: This will complicate things a bit but do we want to support SSML and non-SSML with a flag? The windows provider will not support SSML so we would have to throw an error if that combo is called.

rkusa: All of them support SSML, so everything is SSML right now (even the windows provider). I should prabably mention this in the docs of the `text` field. Eg. I am using tags like `<phoneme alphabet="ipa" ph="w&#618;nd">Wind</phoneme>` in DATIS and they also work for the windows provider.

protos/dcs/tts/v0/tts.proto
rkusa: I derived `wait` from the `--no-wait` flag I often see in CLIs. But I am totally up for another term to make it easier to understand.

I was also considering `blocking`, but avoided it to not confuse it with blocking the DCS mission (doesn't really make sense that this would block DCS, but it came to my mind).

`async` would work for me.

Another idea: get rid of the field altogether and instead return the time in milliseconds it will roughly take to _speak_ the transmission. The client could then sleep that amount. This would have the benefit of avoiding running into request timeouts (in case there are any due to proxies or whatever).

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,94 @@
...
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Tts";
+option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/tts";
+
+service TtsService {
+  rpc Transmit(TransmitRequest) returns (TransmitResponse) {}
+}
+
+message TransmitRequest {
+  // The text that is synthesized to speech and transmitted to SRS.
+  string text = 1;
```
rkusa, line 13: Should we add a second text field for the _plain_ text - i.e. no SSML tags, no transformations like FOUR TREE instead of 43? This plain text could then be emitted as an event for other clients to pickup (eg to present as text to the user for a11y use-cases)?

rurounijones: That is a great idea. Have both SSML and text fields.

protos/dcs/tts/v0/tts.proto
rurounijones: Just to be sure. When I talk about the windows provider I am referring to:


* https://download.voiceelements.com/MsSpeech/SpeechPlatformRuntime.msi
* https://download.voiceelements.com/MsSpeech/MSSpeech_SR_en-US_TELE.msi

Which is not being actively developes but still works for non-SSML use-cases. Are you referring to the same or a different system?

protos/dcs/tts/v0/tts.proto
rurounijones: I think returning the time is a good idea even if we have async option but do all the providers support that?

protos/dcs/tts/v0/tts.proto
rkusa: I am referring to the built-in on Windows and have no ideas what the downloads you've linked are for 😊
This is the one I am talking about: https://learn.microsoft.com/en-us/uwp/api/windows.media.speechsynthesis.speechsynthesizer?view=winrt-22621

protos/dcs/tts/v0/tts.proto
rkusa: I don't get the time from the providers (actually I haven't checked, I'd have just done it as follows), I'd have to make the following changes:
- It always executes the speech synthesizes and opus encoding before the response is send to the client.
- Which is necessary so that I can then estimate the duration based on the amount of opus audio frames.

It looks as follows right now (in the `async` case):

```
Request -> Open Connection to SRS -> Response -> Speech Synthezis -> OPUS encoding -> SRS Transmission
```

To return the Duration, I'd change it to:

```
Request -> Open Connection to SRS -> Speech Synthezis -> OPUS encoding -> Response -> SRS Transmission
```

protos/dcs/tts/v0/tts.proto
rkusa: Giving this a second thought, does this even make sense to have the raw text as an event? E.g. in the use-case of OverlordBot, I think it would make more sense if you execute an `OutText` in parallel right away, instead of relying on the TTS event that contains the raw text. Same goes for DATIS. 🤔

protos/dcs/tts/v0/tts.proto
rurounijones: Hmm. What would the performance impact of that be? I am fine with async being an option but false by default.

protos/dcs/tts/v0/tts.proto
rurounijones: OverlordBot uses the one that requires the download. Not sure what that one is or why I didn't use it if I knew about it. Will have to look when I get back.

protos/dcs/tts/v0/tts.proto
rurounijones: I think there is a benefit to having an emitted event for some special use-cases such as clients for deaf and hard-of-hearing users as well as logging transmissions.

protos/dcs/tts/v0/tts.proto
rkusa: I couldn't find a way to ignore those lines with protolint. I really don't like url shorteners in code docs, I rather see where the link leads right away. How about bumping the max line length from 80 to 100? This is, at least in this case, enough to squeeze in the links and would be consistent with the Rust code (which is also capped at 100).

rurounijones: Increasing to 100 works for me. No point blocking useful stuff for a linter!

src/rpc/tts.rs
rkusa: Created an issue for it #199

protos/dcs/tts/v0/tts.proto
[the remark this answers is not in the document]
rkusa: Moved to config in a1c72a571fcbe3a5d2bae5ccb12e485e26e4e775

protos/dcs/tts/v0/tts.proto
rkusa: The TTS will definitely take a bit longer, as it is waiting for the round-trip to the TTS provider. However, this might be a good thing because it allows to return a possible error from the TTS provider even if `async` is enabled.

1074ee06f64f0fe7b27ca32235f23ee65a693696 renames the field to `async`, and adds the `duration_ms` to the response

protos/dcs/tts/v0/tts.proto
rkusa: Added in cf426c89de55f34943fc4a6070d7c279651e4229

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,82 @@
...
+  // optional, please consider providing it as it can be used to display the spoken text to players
+  // with hearing impairments.
+  optional string text_plain = 2;
+
+  // The radio frequency in Hz the transmission is send to. Example: 251000000
+  // for 251.00MHz.
+  uint64 frequency = 3;
+
+  // Name of the SRS client. Defaults to "DCS-gRPC".
+  optional string name = 4;
```
rurounijones, line 27: Could we add this to the TTS event so that we can format a message like: `251.000MHz | NameOfSender | Message`. May be useful for scripted missions.

And if we do then update the doc to reflect that it will be the SRS client name _and_ the name displayed in events.

protos/dcs/tts/v0/tts.proto
```diff
@@ -0,0 +1,82 @@
+syntax = "proto3";
+package dcs.tts.v0;
+import "dcs/common/v0/common.proto";
+option csharp_namespace = "RurouniJones.Dcs.Grpc.V0.Tts";
+option go_package = "github.com/DCS-gRPC/go-bindings/dcs/v0/tts";
+
+service TtsService {
```
rurounijones, line 7: I wonder if it would be possible to have a `TransmisionInProgress` API call that returns true if _we_ are already transmitting something on a freq. My gut says that this will probably be difficult.

protos/dcs/tts/v0/tts.proto
rkusa: Done, also renamed to `srs_client_name` as it seem to unspecific to me looking at it again.

rurounijones (approved):

rkusa:

PR #196, opened 2022-11-08
Add unit id to GetTransform response

rkusa:
For the [LSO bot](https://github.com/DCS-gRPC/lso), I frequently poll carrier-plane pairs for their transform using the `GetTransform` method. This has the shortcoming, since I query the units by name, that I cannot easily detect whether a unit respawned. It could easily be that a client reselects the same slot over and over again (in between my polling calls), and since the unit name isn't changing, I cannot tell that the underlying unit changed. The easiest solution I could come up with is to include the unit id in the response, so that I can compare the ids between calls to detect a new unit behind the queried name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GetRealTime` API
 - Added `orientation` and `velocity` to `Weapon` object
 - Added DCS `time` of the update to units stream (`StreamUnitsResponse`)
+- Added `id` to `GetTransform` response
 
 ### Changed
 - Unit objects now return the full group object in the `group` field to make event processing easier. This replaces the `group_name` and `group_category` fields and is a backwards incompatible change.
--- a/lua/DCS-gRPC/methods/unit.lua
+++ b/lua/DCS-gRPC/methods/unit.lua
@@ -68,6 +68,7 @@ GRPC.methods.getUnitTransform = function(params)
 
   return GRPC.success({
     time = timer.getTime(),
+    id = tonumber(unit:getID()),
     rawTransform = GRPC.exporters.rawTransform(unit),
   })
 end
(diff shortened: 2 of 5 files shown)


rurounijones:
I am not against this but I am curious if polling `GetTransform` is still needed now that we have transform information in the unit stream. In my mind we could subscribe to the stream and keep an in-memory store of the units and their positions then "poll" that in-memory store like we are polling the DCS-gRPC server itself currently.

The unit stream would then also provide the information that units have been created or destroyed.

rkusa:
> I am not against this but I am curious if polling `GetTransform` is still needed now that we have transform information in the unit stream.

I am avoiding the stream here, as I am really only interested in carriers and carrier-landing capable planes. Which is potentially a very small subset of what the stream would poll for me. So I am avoiding the stream to avoid unnecessary polling of units I don't care about.

> The unit stream would then also provide the information that units have been created or destroyed.

I could also rely on the event stream to detect destroyed/despawned units. However, this would (same as relying on the `UnitGone` of the units stream) require me to employ complex synchronisation between all the concurrent _tasks_ happening in the LSO. E.g. due to scheduling of tasks, the task that watches a plane, could potentially execute the destruction event of a unit after the task that looks for new units already saw the birth event of a new unit with the same name. In other words, I'd have to add complexity to ensure to process the events in order: `birth` → `death` → `birth` → `death` (and not `birth`→ `birth` → `death` → `death`)

The `id` would simplify it down to: if the task that watches a plane sees a changed `id`, it stops and doesn't really care if another task for the same unit name has already be spawned or not.

Another alternative would be to poll `GetUnit` instead of `GetTransform`, as this already includes the id. It is just that `GetTransform` avoids extra data I don't need.

rurounijones:
> looks for new units already saw the birth event of a new unit with the same name. In other words, I'd have to add complexity to ensure to process the events in order: birth → death → birth → death (and not birth→ birth → death → death)

Perhaps I am being thick here but why would you be keying on the name if the ID is available as a truly unique identifier?

rkusa:
> Perhaps I am being thick here but why would you be keying on the name if the ID is available as a truly unique identifier? If we use that then the ordering of events doesn't matter.

Mostly because I cannot get a unit by id. So there is no way to call `GetUnit` or `GetTransform` using the ID (AFAIK). Otherwise I'd most definitely use the ID and rely on the `NotFound` error of the `GetTransform` call.

rkusa:
Reflecting on what I wrote, I'll accept for now that I myself consider it more a band-aid than anything else. I'll instead detect despawned units based on events.

PR #198, opened 2022-12-04
Add `GetBallisticsCount` API.

rurounijones:
Add an API that allows us to count all ballistics objects.

# Note

I am not entirely sure if this is the best way to do it vs having a ballistics count be part of the event stream on a timer but it works and has already fulfilled a [useful purpose](https://forum.dcs.world/topic/308247-invalid-ballistics-objects-being-created-and-not-cleaned-up-resulting-in-fps-impact/page/2/#comment-5095416) so adding it in its current state.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -21,5 +21,6 @@ read_globals = {
   "trigger",
   "Unit",
   "world",
-  "missionCommands"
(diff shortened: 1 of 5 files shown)


rkusa (approved):
LGTM

Issue #200, opened 2022-12-04
Add a TTS API to indicate channel in use


rurounijones:
Follow-up from the main TTS Work.

Add an API that, when called, returns a boolean that indicates if DCS-gRPC is currently transmitting on a certain frequency.
Optionally, add an API that, when called, blocks if DCS-gRPC  is currently transmitting on a certain frequency and doesn't return until the transmission has finished.

These two APIs can be ersatz mutexes to help stop clients stepping over each other on the radios.

PR #202, opened 2022-12-05
Add GRPC.tts(...) Lua API

rkusa:
Add `GRPC.tts(ssml, frequency[, options])` Lua API.

Usage: https://github.com/DCS-gRPC/rust-server/tree/tts-lua#lua-api

The API currently does not provide an means to prevent overlapping transmissions as we don't want to block the mission env when the API is called and because there is no safe way to return the transmission play time back to Lua asynchronously. I'd suggest to solve this via #200 in future.

### DIFF
--- a/protos/dcs/tts/v0/tts.proto
+++ b/protos/dcs/tts/v0/tts.proto
@@ -34,11 +34,11 @@ message TransmitRequest {
 
   // The origin of the transmission. Relevant if the SRS server has "Line of
   // Sight" and/or "Distance Limit" enabled.
-  dcs.common.v0.Position position = 5;
+  Position position = 5;
 
   // The coalition of the transmission. Relevant if the SRS server has "Secure
   // Coalition Radios" enabled. Only Blue and Red are supported, all other
-  // values will fallback to Blue.
+  // values will fallback to Spectator.
   dcs.common.v0.Coalition coalition = 6;
 
   // Whether to keep the request open until the whole transmission was sent. If
@@ -85,4 +85,22 @@ message TransmitRequest {
 message TransmitResponse {
   // The duration in milliseconds it roughly takes to speak the transmission.
   uint32 duration_ms = 1;
+}
+
+/**
+ * Position of an object in DCS
+ *
+ * Latitude and Longitude are in Decimal Degrees format (e.g. 41.33 / 37.21).
+ * Negative values are used for West of the meridian and south of the equator
+ *
+ * Altitude is given in meters above Mean Sea Level (MSL) and can be a decimal
+ * value.
+ */
+message Position {
+  // Latitude in Decimal Degrees format
+  double lat = 1;
+  // Longitude in Decimal Degrees format
+  double lon = 2;
+  // Altitude in Meters above Mean Sea Level (MSL)
+  double alt = 3;
 }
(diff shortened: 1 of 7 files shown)


protos/dcs/tts/v0/tts.proto
```diff
@@ -34,11 +34,11 @@ message TransmitRequest {
 
   // The origin of the transmission. Relevant if the SRS server has "Line of
   // Sight" and/or "Distance Limit" enabled.
-  dcs.common.v0.Position position = 5;
+  Position position = 5;
```
rkusa, line 37: I've noticed that since I've added `u` and `v` to `dcs.common.v0.Position `, it is very inconvenient to use it as an input, which is why I've changed this to a Position that only has `lat`/`lon`/`alt`. Not pretty though 🤷‍♂️

rurounijones: Hmmm, that is a good point. `Position` is used all over the place as it is a built-in DCS'ism... I wonder if we need to rename your enhanced `Position` to something else to differentiate.

rkusa: I think the inclusion of `u`/`v` makes sense for everything we return. So we might have to split it based on input vs output. We could have `dcs.common.v0.Position` for output, and something like `LatLonPosition`/`LatLonPosition`/`InputPosition`/`RequestPosition` as the type for requests?

Unless, we want to provide the choice that position can be provided either via `lat/lon/alt` or via `u/v/alt`. For the Lua API, the latter might be easier to use, as you usually work with `u/v/alt` in the Lua part (that is basically `x/z/y` from the position object you get from DCS).

rurounijones: Option 1: What if we kept a single position but made everything apart from lat/lon/alt optional?
Option 2: I think `InputPosition` as you suggested above is probably the best compromise

rkusa: Option 1: I am not a fan due to:
- requires `null` checks as a client even though it is always included
- could be provided in requests, but wouldn't have an effect

I'll go with Option 2

rkusa: I've implemented Option 2 in [a285bad](https://github.com/DCS-gRPC/rust-server/pull/202/commits/a285bad38d37fdd5a63d757bf5f95149f3f2fd80). It doesn't feel pretty, but the only other option I had was to use a structure similar to (which isn't ideal either; similar to your `DetailedPosition` suggestion, just with reusing a base `Position`):

```proto
message Position {
  double lat = 1;
  double lon = 2;
  double alt = 3;
}

message ExtendedPosition {
  Position position = 1;
  double u = 4;
  double v = 5;
}
```

If you prefer that, I am happy to update the PR again.

rurounijones: The position with u/v is returned so much (All the unit stream events for starters, right?) that I think having the extra `Position` inside of `ExtendedPosition` is a bit icky. I think the option 2 is the best compromise here.

rurounijones (dismissed):
Given the InputPosition change I think it would be worth noting that this is a backwards incompatible change with quite a wide ranging impact.

rurounijones (approved):

Issue #179, opened 2022-09-09
Errors returned to lua


soerenmeier:
When deserialization fails in a `next` call this error gets returned from rust:
https://github.com/DCS-gRPC/rust-server/blob/4a046556392efc7d17ee9d6c630c3fb3957e1359/src/lib.rs#L151-L158
But in lua only the following line get's outputed: `Error retrieving next command: attempt to concatenate a userdata value`. This is very misleading.
I tried to debug this but I haven't found anything. Something might be wrong with the lua version of dcs.
Did this works once and is this a known issue?

mlua:
function that passes the error to lua: [`callback_error`](https://github.com/khvzak/mlua/blob/050ac4f5ad05072faeda9708e82c10156447c629/src/util.rs#L623)
function that converts the mlua error into a string:  [`init_error_registry`](https://github.com/khvzak/mlua/blob/050ac4f5ad05072faeda9708e82c10156447c629/src/util.rs#L850)

rkusa:
This is indeed very unhelpful. I've seen this error too, and as far as I remember, it was caused by the hot-reloading. Are you using the hot-reloading feature by any chance?
It would have been fixed in #150, but we decided to not merge it. If you are using the hot-reloading feature, I am afraid that the only workaround right now would be to not use hot-reloading, or to make a local change similar to: https://github.com/DCS-gRPC/rust-server/pull/150/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R168

soerenmeier:
No, I compiled with `cargo build --release` and then copied dcs_grpc.dll and the pdb file to the mods folder.

rkusa:
I can reproduce it now.

While trying to create a small example that reproduces the error to report the issue to `mlua`, I've learned that this seems to be caused by DCS' `lua.dll`. I couldn't reproduce it outside of DCS, and then dropped DCS' `lua.dll` into my example, and boom, error was there.
I unfortunately have no idea, why this is caused by DCS' `lua.dll` (but I am not surprised either). I've confirmed that DCS runs Lua 5.1.5 and that DCS-gRPC is compiled against exactly this version.

DCS:
![image](https://user-images.githubusercontent.com/409021/194140363-73dae5cb-54c7-4548-b5bb-a8d42be45b17.png)

DCS-gRPC:
https://github.com/DCS-gRPC/rust-server/blob/main/src/lua5.1/include/lua.h#L20

[...]

PR #218, opened 2023-02-23
Add integrity check to validate installation

rkusa:
To help debug issues caused by incomplete installations / updates, this PR adds an integrity check to the Lua files. It works as follows:
- Automatically embed hashes for all Lua files on build-time
- Validate existence and hash of all Lua files on runtime (when server is started)
- Additionally, compare version number of Lua files and DLL and make sure both match.

Example error:

```
2023-02-23 18:48:21.166 ERROR   SCRIPTING (Main): Mission script error: [string "M:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua"]:12: integrity check failed (hash mismatch of `M:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua`), DCS-gRPC is not started, please check your installation
stack traceback:
  [C]: ?
  [C]: in function 'assert'
  [string "M:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc.lua"]:12: in function 'f'
  [string "M:\Development\DCS-gRPC\rust-server\lua\DCS-gRPC\grpc-mission.lua"]:59: in function 'load'
  [string "GRPC.load()"]:1: in main chunk
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,6 +84,10 @@ pub fn start(_: &Lua, config: Config) -> LuaResult<(bool, Option<String>)> {
     log::debug!("Config: {:#?}", config);
 
     if !config.integrity_check_disabled {
+        if env!("CARGO_PKG_VERSION") != config.version {
+            return Ok((false, Some("dcs_grpc.dll version does not match version of DCS-gRPC Lua files; please check your installation!".to_string())));
+        }
+
         if let Err(err) = integrity::check(&config) {
             return Ok((false, Some(err.to_string())));
         }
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -71,16 +72,28 @@ pub fn init(config: &Config) {
 }
 
 #[no_mangle]
-pub fn start(_: &Lua, config: Config) -> LuaResult<()> {
+pub fn start(_: &Lua, config: Config) -> LuaResult<(bool, Option<String>)> {
```
rkusa, line 75: This is the first instance of how I am probably going to fix #179. Instead of relying on the `LuaResult`, return a boolean and an optional error message. The method can then simply be called via `assert(grpc.start())` to raise a Lua error (if the boolean is `false`).

rurounijones (approved):
Looks great. Thank you.

PR #223, opened 2023-04-07
Fix crash of concurrent Win TTS synthesis

rkusa:
Fixes a crash related to Windows' TTS as reported [in Discord](https://discord.com/channels/696721363276267590/962651858760003594/1087405074373955675). This seems to be caused by a certain concurrency of the synthesis. I don't really know the exact cause (could be Windows - Rust interop), but simply preventing concurrency seems to fix it.

### DIFF
--- a/tts/src/win.rs
+++ b/tts/src/win.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 
+use tokio::sync::Mutex;
 use windows::core::HSTRING;
 use windows::Media::SpeechSynthesis::SpeechSynthesizer;
 use windows::Storage::Streams::DataReader;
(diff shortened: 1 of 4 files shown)


tts/src/win.rs
```diff
@@ -9,10 +10,15 @@ pub struct WinConfig {
...
 }
 
+static MUTEX: Mutex<()> = Mutex::const_new(());
+
 pub async fn synthesize(text: &str, config: &WinConfig) -> Result<Vec<Vec<u8>>, WinError> {
     // Note, there does not seem to be a way to explicitly set 16000kHz, 16 audio bits per
     // sample and mono channel.
 
+    // Prevent concurrent Windows TTS synthesis, as this might cause a crash.
+    let lock = MUTEX.lock().await;
```
rkusa, line 20: A semaphore with a reasonable concurrency limit might also do the trick - would would require some additional testing to find a good limit

rurounijones: Lets stop the bleeding with this change and then we can consider setting a concurrency.

I am fine with this but the checks are failing.

rurounijones (dismissed):

uriba107:
Just to chime in.
I've ran one of my Hound demo missions with a DLL built from this branch. the fix seems to have done the trick and my mission was able to run correctly (I was a maximum of 14 messages in 2 seconds, usually two or 3 are the same message sent on different channels).

rkusa:
Rebased to get CI fixes from `main`

rurounijones (dismissed):

rurounijones (approved):

PR #228, opened 2023-06-22
Remove OpenSSL from dependency tree

rkusa:
OpenSSL was not supposed to be in the dependency tree. All dependencies, except rusoto were already configured to use https://github.com/rustls/rustls instead. This PR fixes the oversight and also makes sure that rusoto uses rustls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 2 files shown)


rkusa:
Ignoring the failed Lua CI check, as it is unrelated (probably due to an update of the linter).

PR #231, opened 2023-07-02
Add Coqui (server) as TTS provider

rkusa:
Added [Coqui](https://github.com/coqui-ai/TTS) as another TTS provider (that can be run locally).

It is bummer, that it can not be directly integrated ('cause it's Python based), but offers way better quality than the Windows TTS. See README for instructions.

Sample:
[tts.zip](https://github.com/DCS-gRPC/rust-server/files/11929543/tts.zip)

Should we go ahead and at it even though it requires to run Coqui as a separate server? @rurounijones what do you think?

Alternatives:
https://github.com/rhasspy/piper is the only TTS I know that could be directly integrated into DCS-gRPC eventually (once it builds on Windows and the build steps allow a reasonable integration into `cargo build`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `ActivateGroup` API which allows to activate groups with late activation.
 - Added `DestroyGroup` API which removes the entire group from the game world.
 - Added `DestroyUnit` API
-- Added `GetClients` to `SrsService`, which retrieves a list of units that are connected to SRS and the frequencies they are connected to.
-- Added `SrsConnectEvent` and `SrsDisconnectEvent` events 
+- Added `GetClients` to `SrsService`, which retrieves a list of units that are connected to SRS and the frequencies they are connected to
+- Added `SrsConnectEvent` and `SrsDisconnectEvent` events
+- Added [Coqui](https://github.com/coqui-ai/TTS) as another TTS provider (that can be run locally)
 
 ### Fixed
 - Fixed `MarkAddEvent`, `MarkChangeEvent` and `MarkRemoveEvent` position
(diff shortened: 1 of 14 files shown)


rurounijones:
We already have the windows one as a default and all the other TTS systems, including the cloud ones, are optional based on the desires of the person running the server so I think adding to these options is fine.

The only thing I would be wary about is, as it appears to be in this case, that one TTS does not support SSML. We absolutely have to make that _extremely_ obvious in the documentation to avoid people asking us "Why it doesn't sound right".

rkusa:
@rurounijones very good points! Especially the SSML one I didn't notice - it would be an inconsistency I don't like. Another option I've considered would be to just accept sound files in addition to the existing TTS options (I could then run corqui as part of my client; and we open up DCS-gRPC SRS transmissions for more use-cases in general). What do you think about that as an alternative?

rurounijones:
I think having Coqui as a TTS back-end is still good. Adding a "Play this sound file over SRS" API would _also_ be beneficial but I don't think it needs to be an alternative.

The only questions I have regarding the "play sound file" is how it would work in terms of data transfer. Stream the file to DCS-gRPC and have that stream to SRS? Or upload the file then have DCS-gRPC stream it. If the latter then I would worry about people sending 50MB .wav files and the implications on memory usage and delays in transmissions.

rkusa:
> The only questions I have regarding the "play sound file" is how it would work in terms of data transfer. Stream the file to DCS-gRPC and have that stream to SRS? Or upload the file then have DCS-gRPC stream it? If the latter then I would worry about people sending 50MBaudio files and the implications on memory usage and delays in transmissions.

Latter is basically already happening. Some of the TTS cloud services return the audio as base64 encoded data in a JSON (🤦‍♂️) ... so with a long enough text you can already end up at the same result.

Anyway, I made up my mind about it and rather wait for a Rust native TTS solution. Seeing a big rise of KI stuff in Rust, I expect a native TTS solution soon enough (probably using https://github.com/burn-rs/burn).

PR #232, opened 2023-07-11
add guard in lua against nil group from dead unit

uriba107:
In 2.8.2 ED has changed the behaviour of :getGroup() for units. and it will return nil for dead units.
https://wiki.hoggit.us/view/DCS_func_getGroup

in current version, you would get an empty group and feed it to the exporter class which throws errors in dcs.log
```
2023-06-27 12:52:53.246 ERROR   dcs_grpc: Error in event handler: [string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\exporters\object.lua"]:49: attempt to index local 'group' (a nil value)
stack traceback:
	[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\exporters\object.lua"]:49: in function 'group'
	[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\exporters\object.lua"]:27: in function <[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\exporters\object.lua"]:19>
	(tail call): ?
	[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\methods\mission.lua"]:43: in function 'typed_exporter'
	[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\methods\mission.lua"]:341: in function <[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\methods\mission.lua"]:63>
	(tail call): ?
	[C]: in function 'xpcall'
	[string "C:\Users\Carmen\Saved Games\DCS.openbeta_server\Scripts\DCS-gRPC\grpc.lua"]:233: in function 'onEvent'
	[string "Scripts/World/EventHandlers.lua"]:13: in function <[string "Scripts/World/EventHandlers.lua"]:11>
```

The proposed change tests if the incoming group object is indeed a group (based on lua metatable) and return an empty table if it fails


lua/DCS-gRPC/exporters/object.lua
```diff
@@ -45,6 +45,7 @@ GRPC.exporters.rawTransform = function(object)
 end
 
 GRPC.exporters.group = function(group)
+  if type(group) ~= "table" and getmetatable(group) ~= Group then return {} end
```
rkusa, line 48: I think returning `{}` might lead to effects harder to debug than the current error we get. I haven't tested it, but I'd guess this would simply set all fields of the group to their default / undefined value which is probably unexpected in most cases.

Without giving it too much thought (so please feel free to challenge my suggestion), I think the following might be a better solution:
- explicitly define the [`group` property](https://github.com/DCS-gRPC/rust-server/blob/main/protos/dcs/common/v0/common.proto#L235) as `optional`.
- handle the possibility of `getGroup` returning `nil` [in the unit export](https://github.com/DCS-gRPC/rust-server/blob/main/lua/DCS-gRPC/exporters/object.lua#L27)

Wdyt?

rkusa (commented):
Thanks a lot for the PR, highly appreciated!

lua/DCS-gRPC/exporters/object.lua
uriba107: I initially set it with return nil, but then I suspected that someone might try and iterate over a table. but you are correct. it will force someone to put some guards on his end if.

I'm unfamiliar grpc, but a my gut feeling reading your response is that returning nil in this line, and specifying optional in the proto file is the correct way to go.

I'll make the changes

rurounijones: I _really_ don't like marking the group property as optional as we have now complicated the handling of every client (and will probably break existing clients who are expecting group to always be there).

I haven't heard anyone complaining about missing events (that would have these `nil`s) so my question is.

What is the goal of this change? Just stop the error messages appearing? Publishing essential events that are currently not being published?

rurounijones (changes requested):
See comment @ https://github.com/DCS-gRPC/rust-server/pull/232#discussion_r1262020138

lua/DCS-gRPC/exporters/object.lua
rkusa: Fair point. I wouldn't mind reducing the DeadEvent down to just containing the unit's name instead of a whole unit object. At least in my use-cases, that is all I need to cleanup state about the units I track in my clients. I'd probably even prefer it over a `UnitWithoutGroup`.

rkusa: Also, I wouldn't be one complaining about missing `Dead` events, as I don't trust ED to fire that reliably anyway. So instead, I just consider a unit dead whenever I fetch information for it and receive a not found error 🤪

uriba107: ED is known to make a mess out for things.. for example, if a unit is dead and burning, it still exists as a staticObject until it's blows up. which is why I assume they disconnect it from the group as staticObjects cannot be assigned to groups.

It's super complicated to replicate but AFAIK you might see something like 
```TRUCK1(Unit in CONVOY) --> TRUCK1(burning staticObject) --> TRUCK1(dead Unit)```

that's super tricky and follow the whole thing I'm afraid. which is sometimes the reason you feel the dead events are unreliable.

uriba107: My goal? Not seeing as many error messages in the log.
There is also a slight chance that once an error pops up dcs event handler will. Not continue processing more functions bound to that event but that might be just a feeling

rurounijones: I think the most compatible (But not backwards compatible) solution here is the `DeadUnit` approach.

1. It will get rid of error messages
3. It will give people listening to events the ability to react on this event without needing to consult a local storage of units, which they would otherwise have to do if we just sent the unit ID. This may be important for simple stats trackers.

uriba107: I'm currently trying to debug something on my end. I suspect it has something to do with the grpc throwing LUA errors on parse, but I'll need to be very targeted in my testing, so I'll build a simple test mission tonight and do some tests.

as to how it will be resolved eventually, I trust your judgment for choosing the correct path. but I want to first determine if it's a real issue, as I suspect, or purely cosmetic.

uriba107: Ok, 
I'm circling back to this one. 

I've got a unit testing mission for my project. I had everything green before installing gRPC. 

After compiling from a clean main branch. Tests started failing and the dcs.log was filled with odd nil events from gRPC. Once i re applied this fix, tests are green again. 

Do you have any recommendations about a proper methodology to ensure grpc errors do not break the dcs event handler?

fnslayeroc: whats the word on this? Fox3 has been using gRPC, and anything i dynamically spawn, when it dies triggers this error....

dutchie032: @uriba107  

I don't see how gRPC breaks the events on the next event listener. 
It initially might have, but all events are wrapped in a xpcall function to make sure that if anything goes in the gRPC lua layer it doesn't cascade down. 

Could you specify what events are skipped or have a way to replicate it?

dutchie032: @fnslayeroc  
This error might appear in the logs, but does something functionally break?

PR #239, opened 2023-10-15
Update DCS-gRPC.lua

Applevangelist:
Close file after success to remove lock


rurounijones:
Thanks for the change. Could you update the `CHANGELOG.md` file as well with your change. e.g. In the `FIXED` section of the unreleased version add the fact that the config file is now being closed.

rkusa:
> @rkusa Looks like the codebase is failing the latest version of the linter

#241 👍

rurounijones:
Bypassing protection since the part that is failing is a linter that is fixed elsewhere.

Issue #246, opened 2023-11-17
Error in mission.lua due to DCS API getCategory behaviour change


funkyfranky:
In latest OB patch, the DCS API function getCategory was corrected. It now correctly returns the unit, weapon airbase category depending on the object type the method is used on. Before it always returned the object category.

In the file mission.lua, the lines

`local category = object:getCategory()`

must be changed to

`local category = Object.getCategory(object)`

to correctly retrieve the object category.

Currently, we get the error

> 2023-11-17 01:24:32.044 ERROR   dcs_grpc: Error in event handler: [string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\exporters\object.lua"]:21: attempt to call method 'getID' (a nil value)
> stack traceback:
> 	[string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\exporters\object.lua"]:21: in function <[string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\exporters\object.lua"]:19>
> 	(tail call): ?
> 	[string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\methods\mission.lua"]:73: in function <[string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\methods\mission.lua"]:63>
> 	(tail call): ?
> 	[C]: in function 'xpcall'
> 	[string "C:\Users\frank\Saved Games\DCS.testing\Scripts\DCS-gRPC\grpc.lua"]:233: in function 'onEvent'
> 	[string "Scripts/World/EventHandlers.lua"]:13: in function <[string "Scripts/World/EventHandlers.lua"]:11>

Penecruz:
Also needs to be corrected in unit.lua as well.

peterb154:
Note, I have implemented the changes in this PR locally on top of the 0.7.1 release. https://github.com/DCS-gRPC/rust-server/pull/247 

Even with those fixes in place, the StreamUnit hangs after a unitGone event. 

Here is the output from a little python grpc script streaming units. Whenever a "unit gone" event happens, the stream stops. The unit gone is me jumping out of a slot and taking another slot. It only records the unit gone, no other events..
```
2023-11-19T09:39:34 INFO    :  GRPC is up
2023-11-19T09:39:34 INFO    :  unit     : 3 SSS-Jeep
[...]

PR #247, opened 2023-11-18
Update to new .getCategory

Penecruz:
Update the code to be compatible with the new `.getCategory()` implementation which ED changed in the latest release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed crash of concurrent Windows TTS synthesis ([#223](https://github.com/DCS-gRPC/rust-server/issues/223))
 - Fixed file handler left open: Closing config file again after reading it in LUA
 - Fixed the config value for `tts.defaultProvider` for Google Cloud to be lowercase `gcloud` (instead of `gCloud`)
-
+- Fixed error with getCategory due to API changes in DCS ([#246](https://github.com/DCS-gRPC/rust-server/issues/246))
 ## [0.7.1] - 2023-01-08
 
 ### Fixed
--- a/lua/DCS-gRPC/methods/mission.lua
+++ b/lua/DCS-gRPC/methods/mission.lua
@@ -5,7 +5,7 @@ local function exporter(object)
     return nil
   end
 
-  local category = object:getCategory()
+  local category = Object.getCategory(object) -- change for DCS API fixes in getcategory()
 
   if category == Object.Category.BASE or object.className_ == 'Airbase' then
     -- carriers are of category unit, but are a Airbase class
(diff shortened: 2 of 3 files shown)


peterb154:
Note comment in issue https://github.com/DCS-gRPC/rust-server/issues/246#issuecomment-1817895704

martinco:
I've also been running with the same set of changes as this PR, with the additional change to the reference in `lua/DCS-gRPC/methods/controllers.lua` updated.

I haven't had any issues since, no errors, or stream stopping (I left only the `group:getCategory()` in `exporters/object.lua` for group as i haven't had a chance to test it but with 900 units, and regular destruction and player slot changes on a multiplayer server, no issues since making this set of changes

mobot-gh:
> I haven't had any issues since, no errors, or stream stopping (I left only the `group:getCategory()` in `exporters/object.lua` for group as i haven't had a chance to test it but with 900 units, and regular destruction and player slot changes on a multiplayer server, no issues since making this set of changes

If the reference in `exporters/object.lua` is left as is, the return values will be from the `Group.Category` enum and not the `Object.Category` enum, right? If so, any client code with conditionals based on that value would need to be tweaked unless the line is changed, like the others, to `Object.getCategory(group)`.

martinco:
Sorry for the confusion, all I meant by that anecdote was more the other unit known case and that seems good

But in the group case I haven't checked the behavior/ what it should be and would be guesswork on my part to change without verification/cause  so pointed it out but haven't acted on it 

Apologies for the anecdote; its more that it's not impacting me doesn't mean it's not a problem

mobot-gh:
> Sorry for the confusion, all I meant by that anecdote was more the other unit known case and that seems good
> 
> But in the group case I haven't checked the behavior/ what it should be and would be guesswork on my part to change without verification/cause so pointed it out but haven't acted on it
> 
> Apologies for the anecdote; its more that it's not impacting me doesn't mean it's not a problem

No need to apologise, I added my comment as more of a question to see if it is only units that are affected by the change to `.getCategory()` or if Weapons, Groups, and Statics are also included. Maybe `mygroup:getCategory()` was already returning values from the `Group.Category` enum? Only some more testing will help us understand unless someone knows for certain already.

Penecruz:
Uncertain of other impacts, the changes in this PR have allowed OverlordBot to function without issue, which is my only application using DCS-gRPC. My understanding was only Object was returning incorrect data prior to the change ED made in latest OB.

rurounijones:
Can you rebase this on to the current main branch which should fix lint failures.

mobot-gh:
There is also the reference mentioned by @martinco on [L25 of controllers.lua](https://github.com/DCS-gRPC/rust-server/blob/db29d0248396a25fcfd26cbcce379052db96f11a/lua/DCS-gRPC/methods/controllers.lua#L25)? Should that be included in this PR?

Penecruz:
> There is also the reference mentioned by @martinco on [L25 of controllers.lua](https://github.com/DCS-gRPC/rust-server/blob/db29d0248396a25fcfd26cbcce379052db96f11a/lua/DCS-gRPC/methods/controllers.lua#L25). Should that be included in this PR?

Not needed in this case.

rurounijones (approved):

PR #249, opened 2023-12-06
Filter damaged ships from airbase export

mobot-gh:
Due to the fact that:
1. DCS treats ships as airbases. 
2. DCS removes units from groups when they are critically damaged but not yet destroyed, and changes them to static objects. 

The `getAirbases` method from `WorldService` can fail with an error if it is called while a ship is in this static state, as the exporter attempts to access the group the airbase unit belongs to, which does not exist.

This PR filters out these static ships by performing nil checks for the unit & group before attempting to export.


rurounijones (approved):

rurounijones (changes requested):
Could you add an entry into the CHANGELOG with details of the fix, then we can merge.

mobot-gh:
As `main` has been updated is it cleaner to close this request, rebase, and open a new PR, or is it okay to push a slightly modified history to this PR, or will you resolve it manually (single change in `CHANGELOG.md`)?

rurounijones:
Could you rebase on main and fix the conflict, then I can merge.

rurounijones (approved):

PR #255, opened 2024-01-30
Feature/installer

### LINKED ISSUE #213: Create one-click installer
Create one-click installer that will simplify installation for less technical users.

dutchie032:
fixes #213 

A simple installer for dcs-grpc. 
Version agnostic for simplicity and minimal rebuilding effort. 

Can write some info in main README.md if required, but can also understand if main maintainers wants to update himself after creating the artifact. 

installer/Readme.md:
# Installer of DCS-gRPC

## To Build
`#pyinstaller --clean --onefile  -y -n "DCS-gRPC-Installer" .\main.py`

## What

1. It collects all releases from the Github api.
2. Makes the user select the version (latest probably 99% percent of the time), but fallback possibilities might be nice. 
3. Downloads the .zip artifact and unzips in install location (selected by user)
4. Updates the missionscripting.lua file. 
   - MissionScripting.lua needs to be selected. No other lua file will be accepted as DCS only reads that file.
   - `dofile(lfs.writedir()..[[Scripts\DCS-gRPC\grpc-mission.lua]])` is on the first line below `dofile('Scripts/ScriptingSystem.lua')`
   - all other `dofile(lfs.writedir()..[[Scripts\DCS-gRPC\grpc-mission.lua]])` or similar (grpc-mission.lua) will be removed so in case of reinstall any possible error will be reset

### DIFF
--- a/installer/install.py
+++ b/installer/install.py
@@ -1,0 +1,70 @@
+import os
+import sys
+import shutil
+import zipfile
+import urllib.request
+
+
+def __UpdateMissionScriptingLua(path):
+    gRPCstartLine = "dofile(lfs.writedir()..[[Scripts\DCS-gRPC\grpc-mission.lua]])"
+
+    #Supports replacing grpc-mission.lua to another file if the needs arise later and searching for the new file at the same time
+    searchStrings =  ["grpc-mission.lua"] 
+    searchStringSystem = "scripts/scriptingSystem.lua"
+
+    if not os.path.exists(path):
+        return False
+
+    newlines = []
+    with open(path, 'r') as f:
+        lines = f.readlines()
+        i = 0
+
+        systemLine = -1
+        for line in lines:
+            if any(searchString in line for searchString in searchStrings):
+                newlines += "\n" #ADD EMPTY LINE 
+                line = "\n" #REMOVE LINE
+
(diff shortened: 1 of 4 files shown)


dutchie032:
I saw no CI/CD action that actually build the artifact so kinda assumed it was a manual artifact upload.
Please correct me if I'm wrong and I will also create a CD for publishing the .exe. 

I could only find a ci.yml, but didn't seem to upload the .zip artifact. 
If there is another .yml file, or I overlooked it in ci.yml please let me know

dutchie032:
Solves #213

installer/install.py
```diff
@@ -0,0 +1,70 @@
+import os
+import sys
+import shutil
+import zipfile
+import urllib.request
+
+
+def __UpdateMissionScriptingLua(path):
```
rurounijones, line 8: I may be missing something but won't this add the GRPC line every time the program is run? (e.g. someone using it to update versions)

rurounijones (commented):
Given how this is structured I think this would be better suited to a separate installer repo in the `DCS-gRPC` project group. The advantages being that you can then update it on your own cadence and not require PR reviews from us. When we do a `rust-server` release we can then also include a link to the installer for people who want to use it.

Whaddaya think?

PR #261, opened 2024-05-24
Add Piper as TTS provider

rkusa:
Add [Piper ](https://github.com/rhasspy/piper) as another TTS provider.

Similar to #231, it cannot directly be compiled into DCS-gRPC, but at least doesn't require running an http server. In a nutshell: it requires to manually download Piper, and DCS-gRPC runs the `piper.exe` each time TTS is executed.

I am not sure if I really want to add Piper this way, which is why I'll be keeping this as a draft PR for now.

It definitely has great speech quality and does not require a lot of resources (it is optimized for the Raspberry PIs).

[Sample](https://github.com/DCS-gRPC/rust-server/files/15436084/sample.zip)


rurounijones:
I think having a high quality _local_ TTS provider is worth any dependency niggles like having a separate `exe` (We could always bundle it).

I think merging this is a net positive!.

PR #269, opened 2024-09-11
Add API Key Authentication

dutchie032:
Added API Key authentication. 

Due to a recurring question about authentication of clients I've implemented a Interceptor layer to the tonic server to check all calls for valid api keys. 

example config: 
```
auth.enabled = false --defaults to false
auth.tokens = {
    { client = "test", token = "Sometest" },
    { client = "another client", token = "Some other test" }
}
```
`auth.tokens` is a table of auth keys with their client name. 
There can be a "default" client or a single key for all clients, but this is up to configuration. 
There can be as many client keys as needed. 

In the debug log the client that authenticates is logged. 

### Performance
Performance wise there is no notable difference. 

### Possible future features
Possibly in the future "expiration_date" can be added to automatically revoke issues, but I didn't think that was needed for a first implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,15 @@ throughputLimit = 600
 -- Whether the integrity check, meant to spot installation issues, is disabled.
 integrityCheckDisabled = false
 
+-- Whether or not authentication is required
+auth.enabled = false 
+-- Authentication tokens table with client names and their tokens for split tokens. 
+auth.tokens = {
+  -- client => clientName, token => Any token. Advice to use UTF-8 only. Length not limited explicitly 
+  { client = "SomeClient", token = "SomeToken" }, 
+  { client = "SomeClient2", token = "SomeOtherToken" }
+}
+
 -- The default TTS provider to use if a TTS request does not explicitly specify another one.
 tts.defaultProvider = "win"
 
--- a/README.md
+++ b/README.md
@@ -226,9 +226,9 @@ In order to develop clients for `DCS-gRPC` you must be familiar with gRPC concep
 
 The gRPC .proto files are available in the `Docs/DCS-gRPC` folder and also available in the Github repo
 
-### Authentication
+### Client Authentication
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -81,6 +81,15 @@ throughputLimit = 600
 -- Whether the integrity check, meant to spot installation issues, is disabled.
 integrityCheckDisabled = false
 
+-- Whether or not authentication is required
+auth.enabled = false 
+-- Authentication tokens table with client names and their tokens for split tokens. 
+auth.tokens = {
+  -- client => clientName, token => Any token. Advice to use UTF-8 only. Length not limited explicitly
```
rurounijones, line 88: There is an `eval` feature controlled in the config that allows for callers to execute arbitrary code, if it isn't too much effort then adding that to the hash and controlling for it would be very useful I think

```lua
{ client = "test", token = "Sometest", eval = true },
{ client = "test", token = "Sometest", eval = false },
```

rurounijones (commented):
Can you update the README with a section specifically on Authentication with information on what clients should do (In general terms) to add authentication to their requests.

README.md
dutchie032: Looked into this. 
I like the idea as well, but from a Request<Body> I don't seem to easily receive the called service/method. 
I definitely want to have another go at it later, so maybe a seperate issue for it would be good. 
Maybe someone else does known.

I didn't really want to check the metadata twice (once in the Interceptor and once in the function `eval` itself. 
Not sure if it's due to the version (that uses the generic `http::request::Request` and not `tonic::Request` which might have some more options. 

So if it's alright with you I'd like to suggest to mark this suggestion as a seperate improvement :)

dutchie032:
Updated PR with example c# code. 
It's collapsible so won't take too much space, but a dev can have a look at the code. 
(I know there's Go clients and maybe even others, but I didn't want to add code that I didn't know to work)
Would be nice to have it added by someone if they implement it in Go or other languages.

rurounijones (approved):

PR #296, opened 2026-05-04
Change argument order for atan2 call in atmosphere.rs

mobot-gh:
The output from the getWind rpc wasn't matching static wind set in missions. By reversing the argument order in the atan2 method in the get_wind_heading_and_strength function in atmosphere.rs, the correct output is achieved.


YoloWingPixie (commented):
LGTM

- Don't forget to update CHANGELOG unreleased section

YoloWingPixie (approved):
LGTM

PR #297, opened 2026-05-04
Add getUnitAgl rpc and update streamunits with same

mobot-gh:


YoloWingPixie (commented):
LGTM

- Don't forget to update CHANGELOG Unreleased section

PR #298, opened 2026-05-04
Add coalition, toAll and playerName to sendchat event

mobot-gh:
this allows for discord integrations to pipe game chats to the correct channels and add coalition colouring via ansi formatting

### DIFF
--- a/lua/DCS-gRPC/grpc-hook.lua
+++ b/lua/DCS-gRPC/grpc-hook.lua
@@ -45,19 +45,25 @@ end
 -- None of these methods should return anything as doing so breaks other scripts attempting to
 -- react to the hook as well.
 
-function handler.onPlayerTrySendChat(playerID, msg)
+function handler.onPlayerTrySendChat(playerID, msg, all)
   -- note: currently `all` (third parameter) will always `=true` regardless if the target is to the coalition/team
   --        or to everybody. When ED fixes this, implementation should determine the dcs.common.v0.Coalition
-
+  local coalition = net.get_player_info(playerID, "side")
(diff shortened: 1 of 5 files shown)


lua/DCS-gRPC/grpc-hook.lua
```diff
@@ -45,19 +45,25 @@ end
 -- None of these methods should return anything as doing so breaks other scripts attempting to
 -- react to the hook as well.
 
-function handler.onPlayerTrySendChat(playerID, msg)
+function handler.onPlayerTrySendChat(playerID, msg, all)
   -- note: currently `all` (third parameter) will always `=true` regardless if the target is to the coalition/team
   --        or to everybody. When ED fixes this, implementation should determine the dcs.common.v0.Coalition
-
+  local coalition = net.get_player_info(playerID, "side")
+  local playerName = net.get_player_info(playerID, "name")
```
YoloWingPixie: nit: this could theoretically fail if somehow the player disconnects around the same time as a chat is sent. (It's very unlikely).

Personally I would wrap any call like this in a pcall with a safe default applied if the call fails. However, I acknowledge that this is a far edge case and I don't believe all of our DCS MSE API calls are safe as is.

YoloWingPixie (approved):
LGTM

I left a nit to consider, but I am ok merging as is.

lua/DCS-gRPC/grpc-hook.lua
mobot-gh: Added a nil guard

YoloWingPixie (dismissed):
LGTM

dutchie031 (approved):

PR #299, opened 2026-05-05
Api compatibility improvements

mobot-gh:
this adds some nil guards and other protections required due to changes in how diligently DCS sticks to it's own rules about objects and their properties. log bloat is substantially reduced.

### DIFF
--- a/lua/DCS-gRPC/exporters/object.lua
+++ b/lua/DCS-gRPC/exporters/object.lua
@@ -59,7 +59,7 @@ end
 
 GRPC.exporters.weapon = function(weapon)
   return {
-    id = weapon:tonumber(),
+    id = weapon.id_,
     type = weapon:getTypeName(),
     rawTransform = GRPC.exporters.rawTransform(weapon),
   }
@@ -110,7 +110,9 @@ end
 -- https://wiki.hoggitworld.com/view/DCS_Class_Object
 GRPC.exporters.unknown = function(object)
   if not object.getName then
-    return {}
+    return {
+      name = tostring(object.id_),
+    }
   end
   return {
     name = tostring(object:getName()),
--- a/lua/DCS-gRPC/exporters/object.lua
+++ b/lua/DCS-gRPC/exporters/object.lua
@@ -17,6 +17,10 @@ GRPC.exporters.position = function(pos)
 end
 
 GRPC.exporters.unit = function(unit)
+  local locgroup = Unit.getGroup(unit)
+  if locgroup then
+    locgroup = GRPC.exporters.group(locgroup)
+  end
   return {
     id = tonumber(unit:getID()),
     name = unit:getName(),
@@ -24,7 +28,7 @@ GRPC.exporters.unit = function(unit)
     coalition = unit:getCoalition() + 1, -- Increment for non zero-indexed gRPC enum
     type = unit:getTypeName(),
     playerName = Unit.getPlayerName(unit),
-    group = GRPC.exporters.group(Unit.getGroup(unit)),
+    group = locgroup,
     numberInGroup = unit:getNumber(),
     rawTransform = GRPC.exporters.rawTransform(unit),
   }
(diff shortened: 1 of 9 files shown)


lua/DCS-gRPC/exporters/object.lua
```diff
@@ -55,7 +59,7 @@ end
 
 GRPC.exporters.weapon = function(weapon)
   return {
-    id = tonumber(weapon:getName()),
+    id = weapon:tonumber(),
```
YoloWingPixie: I need to check on this but I wasn't aware that `:tonumber()` is a valid method of DCS objects?

lua/DCS-gRPC/exporters/object.lua
```diff
@@ -105,6 +109,9 @@ end
 -- in the base object of the hierarchy
 -- https://wiki.hoggitworld.com/view/DCS_Class_Object
 GRPC.exporters.unknown = function(object)
+  if not object.getName then
+    return {}
```
YoloWingPixie: The `Unknown` message type has a non-optional string `name`:

https://github.com/DCS-gRPC/rust-server/blob/86c360f41104dc4258300e26b9d45fd602f057e7/protos/dcs/common/v0/common.proto#L204-L206

So I believe this would cause a deserialization error, which I don't think we would want.

YoloWingPixie (changes requested):
I *could* be wrong, but left 2 comments to consider

lua/DCS-gRPC/exporters/object.lua
mobot-gh: I think I picked that code snippet up from another DCS-gRPC user somewhere but it does return a usable value.
Output from `MissionService.streamEvents` taken from Kreya:
```lua
"weapon": {
      "id": 20614657,
      "type": "TOW2",
      "position": {
        "lat": 50.74161157133745,
        "lon": 9.053334613963312,
        "alt": 328.22222900390625,
        "u": -806260.5797191076,
        "v": -372148.9826882064
 },
 ```
 I'd guess that either ED added an undocumented `tonumber()` method that returns some numeric property of the weapon (likely to be `id_`), or the global `tonumber()` function is still in scope when called by `weapon.tonumber(weapon)` and it returns the first numeric property of the weapon object that it finds (also probably `id_`).
 
If the above feels too vague, it could be changed to `id = weapon.id_` as every object in the mission _should_ be  given a runtimeID referenced by their `id_` property.

File: lua/DCS-gRPC/exporters/object.lua (hunk 2 above)
Comment (mobot-gh) [reply lua/DCS-gRPC/exporters/object.lua]:
```lua
if not object.getName then
    return {
      name = '',
    }
end
```

or

```lua
if not object.getName then
    return {
      name = 'UnnamedObject',
    }
end
```

Any preference?

lua/DCS-gRPC/exporters/object.lua
dutchie031: I am on the same boat as to the vagueness. 
I feel indeed either using the ["id_"] might be cleanest? Should be unique and I know almost everything has one. 

:tonumber() might indeed be an undocumented method, which makes it somewhat equal to [id_] as both aren't exactly documented. 

Might be best with `["id_"]` and then have an additional field with "name"? 
Or will name always return `["id_"]` for weapon objects? 
In any case, I feel more comfortable with ["id_"] than `:tonumber()` mostly cause I've seen [id_] being present for a lot of objects.

lua/DCS-gRPC/exporters/object.lua
dutchie031: None from me, maybe @YoloWingPixie Otherwise pick whatever you think is best in this flow. 
(can also again pick the id_ as the name when there is `getName`, however that feels like a flawed direction as well)

lua/DCS-gRPC/exporters/object.lua
YoloWingPixie: I'm going to say that I also feel most comfortable with `id_` since that's well established in the DCS MSE.

lua/DCS-gRPC/exporters/object.lua
YoloWingPixie: I'm in the camp of using id if we can't get the name. It's flawed but it provided more specific information than just a generic name, we should have access to it if we have the object, and provides slightly more context to developers if we fail to get a name.

YoloWingPixie (approved):
LGTM

